PR #319, opened 2013-08-15
Process archived logs in datestamped directories

rgabo:
A new `--process-archive` option of emr-etl-runner allows to process each datestamped directory under the archive bucket, specified by a required `--start` and an optional `--end` argument (if `--end` is omitted, the current day is used).

It does so by creating separate filecrush steps for each datestamped directory without having to move logfiles to the processing bucket. This saves hundreds of thousands of S3 requests and even more time.

If a datestamped directory between the start and end date is missing, the job flow will fail at that particular step so it's by no means foolproof.

### DIFF
--- a/3-enrich/emr-etl-runner/config/config.yml.sample
+++ b/3-enrich/emr-etl-runner/config/config.yml.sample
@@ -28,7 +28,7 @@
     :task_instance_bid: 0.015 # In USD. Adjust bid, or leave blank for non-spot-priced (i.e. on-demand) task instances
 :etl:
   :job_name: SnowPlow ETL # Give your job a name
-  :hadoop_etl_version: 0.4.0 # Version of the Hadoop Enrichment process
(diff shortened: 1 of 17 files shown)


alexanderdean:
@rgabo - it looks like the new `--srcPrefixesFile` in S3DistCp in AMI 3.8 might be a better solution than this PR now? http://docs.aws.amazon.com/ElasticMapReduce/latest/DeveloperGuide/UsingEMR_s3distcp.html#UsingEMR_s3distcp.options

rgabo:
@alexanderdean yes, this is pretty cool and all files will have the right relative path (`./log.gz`).

alexanderdean:
Okay great - I am going to close this PR... Note that r66 Oriental Skylark is AMI 3.6+ (Hadoop 2.4).

alexanderdean:
@rgabo - we need to check how this new option handles conflicts eg:

```
prefix-1/part-00001
prefix-2/part-00002
```

PR #1173, opened 2014-11-12
EmrEtlRunner: enable running on an existing EMR jobflow

rupeshmane:
Pass only emr/jobflow_id as parameter in config file to specify the cluster.


alexanderdean:
Moving to Persistent EMR cluster milestone

alexanderdean:
This is still a neat idea @rupeshmane but we are gradually moving to Dataflow Runner, which already supports targeting an existing EMR cluster. So this is a won't merge.

PR #1360, opened 2015-01-27
Scala Stream Collector: add healthcheck for load balancers

duncan:
We’re deploying a test instance of the scala snowplow collector behind an Amazon Elastic Load Balancer and needed a path that would return a 200 OK response. This adds a `/health` URI path that does just that with a minimum of fuss. I pulled this off the `core-2015-refresh` branch as it looked to be where the latest primary dev was happening. I’m happy to change that and pull from some other branch, if more appropriate. 

An alternate way of accomplishing the same thing would be for `/` to return a 200 OK with a `nothing to here` reply.

### DIFF
--- a/2-collectors/scala-stream-collector/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
+++ b/2-collectors/scala-stream-collector/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
@@ -1,4 +1,4 @@
-/* 
+/*
  * Copyright (c) 2013-2014 Snowplow Analytics Ltd. All rights reserved.
  *
  * This program is licensed to you under the Apache License Version 2.0, and
@@ -134,6 +134,11 @@ class CollectorService(
             }
           }
         }
+      }
+    } ~
+    get {
+      path("health".r) { path =>
+        complete(responseHandler.healthy)
       }
     } ~
     get {
--- a/2-collectors/scala-stream-collector/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/ResponseHandler.scala
+++ b/2-collectors/scala-stream-collector/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/ResponseHandler.scala
@@ -1,4 +1,4 @@
-/* 
+/*
  * Copyright (c) 2013-2014 Snowplow Analytics Ltd. All rights reserved.
  *
  * This program is licensed to you under the Apache License Version 2.0, and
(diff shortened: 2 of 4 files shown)


alexanderdean:
Hey @duncan! This looks cool, thanks for the contrib. Have you signed our CLA - https://github.com/snowplow/snowplow/wiki/CLA

This will get scheduled into Kinesis #5 which will add a lot of bells and whistles to the Kinesis Enrich...

alexanderdean:
Hey @pkallos - how come you didn't need this already? I thought you were running behind ELB...

duncan:
@alexanderdean I haven't yet signed a CLA. Checking with our peeps to see if I can do that or if one of our officers need to do it.

Also, do y'all have any plans to pull the Scala collector out into its own repo? To deploy into our infra, we build from source and pulling in all of `snowplow/snowplow` and what not is a bit of a juggle. I don’t mind it, but if it’s going to go into its own repo—like the trackers—I won’t put the time in on our side to work about it.

alexanderdean:
Cheers Duncan. No plans currently to pull anything other than the trackers out into their own repos - it would be just an unmanageable number of repos and git submodules (we are adding new apps into the repo at a steady rate - two more Kinesis apps coming shortly). Have you looked at Peru? It's pretty good at pulling down subsets of repos... https://github.com/buildinspace/peru

duncan:
Ok, CLA is signed, so that should be good to go. As to the sub repos—yeah. Dealing with submodules is a royal pain, so I’m not actively arguing that you *should* do that... :smile: Just good to know that the things we need to do on our side are worthy to do now rather than later. Thanks! And thanks for the link to Peru. I'll give it a look.

alexanderdean:
CLA confirmed! Yes, the most annoying thing about the single repo approach is that Travis _still_ doesn't have proper testing for multiple projects inside the same repo. But yes 25 more repos in the snowplow org is a worse nightmare...

alexanderdean:
Bringing this forward...

alexanderdean:
@duncan - we've brought this forward into r60, so should be out in a couple of days...

pkallos:
> Hey @pkallos - how come you didn't need this already? I thought you were running behind ELB...

I just used `/i` as healthcheck :).

alexanderdean:
That was my guess ;-)

alexanderdean:
This is done, closing.

duncan:
I considered using `/i` as the health check but didn't want to generate events off that. :) Thanks @alexanderdean for bringing it forward and putting it in the release. I'll rebase our install!

alexanderdean:
Yes, healthcheck events to /i will end up in the bad bucket as querystring-less, but agree it's just another thing to debug...

pkallos:
true, healthcheck is the way to go I was just lazy/in a hurry :+1:

PR #1367, opened 2015-01-30
Scala Kinesis Enrich: unified logger configuration

kazjote:
## Problem

I can't configure logging level of Scala Kinesis Enricher. I can configure the level for Snowplow part of the app but not AWS libraries.

## Solution

Including log4j-over-slf4j unifies the way of configuring logging for both Snowplow part and AWS library. For example, I can decrease the level of logging with: ```-Dorg.slf4j.simpleLogger.defaultLogLevel=warn```


alexanderdean:
Makes sense! Thanks for the contrib. Can't see you guys in the CLA yet - https://github.com/snowplow/snowplow/wiki/CLA

kazjote:
I have just signed CLA - thanks!

alexanderdean:
Confirmed @kazjote - thanks so much!

PR #1371, opened 2015-02-01
Release/r60 bee hummingbird

alexanderdean:


2-collectors/thrift-schemas/collector-payload-1/src/main/thrift/collector-payload.thrift
```diff
@@ -0,0 +1,33 @@
...
+  50: i64 timestamp
+  60: string ipAddress
+  70: optional string userAgent
+  80: optional string refererUri
+  90: optional list<string> headers
+  100: optional string networkUserId
+  110: string path
+  120: optional string contentType
+  130: optional string body
+}
```
alexanderdean: Hey @fblundun - can we come up with a more human-readable ordering/numbering/grouping of these fields please. Anyone coming to this cold won't be able to discern any pattern to this structure... A more coherent ordering would look something like:

    Required fields which are intrinsic properties of HTTP
    Required fields which are Snowplow-specific
    Optional fields which are intrinsic properties of HTTP
    Optional fields which are Snowplow-specific

It would be worth giving each of these four clusters their own block (100, 200, 300, 400 kind of thing).

Also - it seems odd that path is required. I think path always being set is just an artifact of the Scala Stream Collector implementation - certainly path can be null in our CloudFront and Clojure Collectors.

fblundun: I see what you mean. I've made a new branch with a hopefully more readable version [here](https://github.com/snowplow/snowplow/commit/5fdbd597cb1954dd6a4e9049f259bb81235e3071).

It looks like the AdapterRegistry rejects anything without an approved path [here](https://github.com/snowplow/snowplow/blob/master/3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/AdapterRegistry.scala#L77) - wouldn't this reject all events from the Clojure and Cloudfront Collectors with a null path?

PR #1414, opened 2015-02-12
Feature/cloudfront access log adapter

fblundun:


fblundun:
Closing because there are a couple of fields I need to convert from string to number...

fblundun:
Reopening now that the generated JSON matches the schema.

alexanderdean:
I don't think this can be reopened...

alexanderdean:
Nice! Is there a JSON Path file in there somewhere too...

fblundun:
Just assigned myself, I'll write it now...

PR #1416, opened 2015-02-12
Feature/cloudfront access log adapter

fblundun:


alexanderdean:
This is looking good! Something I remembered - CloudFront access log files start with two header rows:

```
#Version: 1.0
#Fields: date time x-edge-location sc-bytes c-ip cs-method cs(Host) cs-uri-stem sc-status cs(Referer) cs(User-Agent) cs-uri-query cs(Cookie) x-edge-result-type x-edge-request-id x-host-header cs-protocol cs-bytes time-taken 
```

It would be good if these were swallowed (=> None) rather than turned into bad rows...

fblundun:
Good point! I'll implement this in the loader rather than the adapter (since adapters have to return NonEmptyLists of raw events)

alexanderdean:
Thanks Fred...

alexanderdean:
Also note that the exact contents of line 2 will be somewhat different across the different access log formats.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/CloudfrontAccessLogAdapter.scala
```diff
@@ -0,0 +1,181 @@
...
+ * Transforms a Cloudfront access log into raw events
+ */
+object CloudfrontAccessLogAdapter {
+
+  /**
+   * Adapter for Cloudfront web distribution access log files
+   */
+  object WebDistribution extends Adapter {
+
+    private val fieldNames = List(
```
alexanderdean: Capital f for FieldNames please!

alexanderdean:
Looks great Fred! Beyond the comment about an upper-case f, can you do two things:

* Create a ticket for the Hadoop test you added
* Remove any CHANGELOG changes to prevent merge conflicts when I pull into the release branch

Once those 3 things are done, I will merge into pygmy-parrot...

fblundun:
I moved the commits onto the latest version of release/r61-pygmy-parrot and now the CHANGELOG is unchanged.

alexanderdean:
Nice - thanks Fred!

PR #1429, opened 2015-02-17
Try to use Net-a-porter parser if httpclient fails

danisola:
PR to use the Net-a-porter parser in case URLEncodedUtils.parse() fails


alexanderdean:
Cool! Scheduling...

PR #1599, opened 2015-04-10
Don't die when checking the cluster and the connection times out

danisola:
Adding a new exception to recover from in `EmrJob.wait_for()`. We've stumbled upon this one last night.


alexanderdean:
Oh nice! Putting into the next main EmrEtlRunner release...

Issue #1643, opened 2015-04-28
Scala Common Enrich: box mkt_term, mkt_content in ME.toTsvSafe


alexanderdean:
/cc @androni, #1630

andrioni:
It also happens with `mkt_content`, other fields might be affected.

alexanderdean:
Duplicate of #1710

PR #1710, opened 2015-05-23
Scala Common Enrich: make mkt_ and refr_ fields TSV safe

jasonbosco:
Currently, the presence of a `%09` (tab) character in the `mkt_*` or `refr_*` fields causes an extra `\t` to be inserted in the enriched data when `%09` is decoded and this causes issues with the TSV format.

This PR sanitizes these fields to make them TSV safe.

Related to issue: #1643


alexanderdean:
Many thanks @jasonbosco ! CLA signed; this is going into the next release...

jasonbosco:
Great! Thanks @alexanderdean

alexanderdean:
This is done, closing!

PR #1927, opened 2015-07-27
Kinesis redshift dripfeed

denismo:
Hello,

please take a look at our draft implementation of the Kinesis Dripfeeder, that we blogged about a few days ago.

Apart from the information that is in the blog and in the README for the project, there is also one important bit that you need to be aware of. There was a bit of logic in Shredder which produced JsonSchemaPairs with validation messages. For our logic to work, I needed the JsonSchemaPairs with individual error messages (or just the valid ones), but it was packaging all of Json Schemas together from one input Json if any of them had an error. So that resulted in the changes ShredJob and Shredder which you may want to look at as I had to remove the validation altogether.

Changes to 1-trackers/\* are accidental but don't know how to revert them.

Denis


alexanderdean:
Huge thanks for the contribution @denismo ! Please can DigDeep sign our CLA:

https://github.com/snowplow/snowplow/wiki/CLA

Scheduling...

jramos:
I can get this to compile and run with `sbt`, but I can't get it to assemble as a runnable JAR.

    $ ./target/scala-2.10/kinesis-redshift-sink-0.1.0 
    Error: Invalid or corrupt jarfile ./target/scala-2.10/kinesis-redshift-sink-0.1.0

Any ideas?

jramos:
Unzipping the assembly gives me this:

    $ unzip kinesis-redshift-sink-0.1.0
    Archive:  kinesis-redshift-sink-0.1.0
    error: End-of-centdir-64 signature not where expected (prepended bytes?)
      (attempting to process anyway)
    warning [kinesis-redshift-sink-0.1.0]:  43 extra bytes at beginning or within zipfile
      (attempting to process anyway)
    ...

denismo:
Hi Justin,

don't know why zip gives these errors. I did have a problem with JAR file
initially - there was no Main-Class entry in the MANIFEST so it was not
self-executable. It should have been fixed though it's worth checking.

Denis

denismo:
BTW, unfortunately while using the drip-feeder in production we identified
some performance problems with it. It cannot insert the data at the rates
that would be sufficient for typical website loads. The fastest that we
could get it to run is 15 events/sec. So we'd advise to hold off using it
in any production system until further update.

Denis

On Fri, Sep 4, 2015 at 9:51 AM, Denis Mikhalkin <denis@digdeepdigital.com.au

mark9white:
We have found the same performance issue - the feeder needs to be changed to upload to S3 then load from that into Redshift rather than using JDBC.

We've made fixes to enable it to be run as a JAR and a few other things at: https://github.com/myinsiders/snowplow

denismo:
Cool, thanks. I've been experimenting with the lighter COPY-over-SSH
approach (to avoid polluting S3 with temp files), with dynamic batching to
ensure predictable load. It is in POC right now, hopefully will have some
time to productize it soon.

Denis

jramos:
Before this can be used with the 0.8.0 Redshift schema, the following change needs to be made:

https://github.com/jramos/snowplow/commit/b1d779aff8d690b2cc85fc3ca8ca6d5cd05587d5

jramos:
I also noticed one issue: when a COPY errors, for example when the fields are mapped to an incorrect type, Redshift reports that the load was a "success". You have to drill into the load details in the console to see that there were elements that didn't COPY over. Not sure of the best way to handle this yet...

jramos:
Ah, I think `MAXERROR 1000` is to blame.

denismo:
1. produceCombinedContext is pretty important, so even with 0.8.0 it is required.

2. Yes, MAXERROR is intentional - we don't want to abort the transaction just because there is an error  in some schema. Those errors do happen - e.g. often when new schemas are added. Because dripfeeding is a continuous process the new schemas are picked up immediately but often the first few events are the test events (and tests fail) so this was introduced. 

In a typical client, you might have 5 to 10 schemas per transaction, and because this is not a batch (with possible manual recovery) process, it is better to produce stl_load_errors for some schemas, while allowing all other schemas (and the main table) to go in. At least it was so for our clients. The limit should of course be configurable though.

jramos:
I don't think 0.8.0 and produceCombinedContext will work together. The schema changes assume all contexts are shredded and `context` will be null.

aldrinleal:
@jramos, you must use:

```
$ java -cp <path to jar> com.snowplowanalytics.snowplow.storage.kinesis.redshift.SinkApp --config <config>
```

Or switch to Java 8. Its a limitation of older vms, where the MANIFEST wound't be read if past 65K files in the .jar file.

Link: http://stackoverflow.com/a/27250943/39261

aldrinleal:
@denismo, I found a small gotcha: When dedup is off (I do when debugging), events simply doesn't get logged. Here's how I solved it (as a patch, sorry):

```
diff --git a/4-storage/kinesis-redshift-sink/src/main/scala/com.snowplowanalyti
index a150709..a4129ce 100644
--- a/4-storage/kinesis-redshift-sink/src/main/scala/com.snowplowanalytics.snow
+++ b/4-storage/kinesis-redshift-sink/src/main/scala/com.snowplowanalytics.snow
@@ -13,6 +13,7 @@
 package com.snowplowanalytics.snowplow.storage.kinesis.redshift
 
 import java.sql.BatchUpdateException
+import java.util
 import java.util.Properties
 
 import com.snowplowanalytics.snowplow.storage.kinesis.redshift.limiter.RatioFl
@@ -72,11 +73,9 @@ class RedshiftEmitter(config: KinesisConnectorConfiguration,
     RatioFlushLimiter.totalKinesisRecords.addAndGet(buffer.getRecords.size)
     val errors: mutable.MutableList[EmitterInput] = new mutable.MutableList[Em
     try {
-      if (deduplicate) {
-        Deduplicator.deduplicate(buffer.getRecords)
-      } else {
-        buffer.getRecords
-      }.foreach { record =>
+      val source: Iterable[EmitterInput] = if (deduplicate) Deduplicator.dedup
+
+      source.foreach { record =>
         try {
           shredder.shred(record._1)
         }
```

cdimitroulas:
is this still being worked on and if so what is the current status?

jramos:
I have a version I'm updating to work with r88. I will comment here when it's ready.

alexanderdean:
No, this is pretty stale now. Snowplow R89 (the next release) will port Hadoop Shred to Spark Shred. A release shortly after that will rewrite StorageLoader into Scala. Then we'll merge the two together, and after that we'll add a Kinesis source to the merged app. That's our dripfeeding strategy for Redshift - closing this PR...

PR #1945, opened 2015-08-04
Upgrade default instance types in emr-etl-runner sample config

danisola:
Instance type m1.small was only available to Hadoop 1, so with the 3.6.0
AMIs the instance type needs to be updated.

See: http://docs.aws.amazon.com/ElasticMapReduce/latest/DeveloperGuide/emr-plan-ec2-instances.html


alexanderdean:
This has been done in R70: https://github.com/snowplow/snowplow/blob/feature/r70/3-enrich/emr-etl-runner/config/config.yml.sample#L40

Issue #1802, opened 2015-06-23
Postgres: add event_vendor/name/format/version to atomic.events


alexanderdean:
Sister ticket of #1801

Issue #1801, opened 2015-06-23
Redshift: add event_vendor/name/format/version to atomic.events


alexanderdean:
@danisola will work on this

pvdb:
> `:g/will/s/will/might/g`

:smile:

Issue #1800, opened 2015-06-23
Scala Common Enrich: write out event_vendor/name/format/version fields


alexanderdean:
I think the most futureproof way of doing this would be to have a structure in atomic.events like this:

```
event       | event_vendor                   | event_name  | event_format | event_version
-----------------------------------------------------------------------------------------
unstruct    | com.snowplowanalytics.snowplow | ad_click    | jsonschema   | 1-0-0
page_view   | com.snowplowanalytics.snowplow | page_view   | jsonschema   | 1-0-0
struct      | com.google.analytics           | event       | jsonschema   | 1-0-0
transaction | com.snowplowanalytics.snowplow | transaction | jsonschema   | 1-0-0
```

The rationale here is:

* Anybody using the existing event column can continue to do so
* Anybody using unstructured events can use the new event_vendor/name/format/version columns, which map directly onto the Iglu URI components
* As we convert our legacy events (e.g. GA-style 5-field structured events) into JSONs, then the corresponding event_vendor/name/format/version columns are futureproofed - their values won't change

Depends on #1637

alexanderdean:
@danisola will work on this

pvdb:
> `:g/will/s/will/might/g`

:smile:

Issue #1637, opened 2015-04-24
Scala Common Enrich: port JSON Schema validation from Scala Hadoop Shred


alexanderdean:
This will validate:

1. unstructured events
2. custom contexts

Note that Scala Hadoop Shred will retain its existing validation code too (i.e., we won't assume an enriched event is valid).

alexanderdean:
@danisola will work on this.

pvdb:
> `:g/will/s/will/might/g`

:smile:

alexanderdean:
From an architecture perspective, my suggestion is to:

1. Extract the current JSON Validation logic from Scala Hadoop Shred into Common Enrich
2. Use the Common Enrich's JSON Validation logic in Scala Hadoop Shred, Scala Hadoop Enrich and Scala Kinesis Enrich

alexanderdean:
Duplicate of #1947

PR #1947, opened 2015-08-06
Scala Common Enrich: port JSON Schema validation from Scala Hadoop Shred

danisola:
This PR is a first attempt to solve https://github.com/snowplow/snowplow/issues/1637.

It moves `Shredder` from Scala Hadoop Shred to Scala Common Enrich. I've just put it in the `utils` package, but I can move it to `utils.shredder` if you want. It also adds context and unstructured event validation in `EnrichmentManager`.

Once we're happy with the changes, I'll proceed to work on https://github.com/snowplow/snowplow/issues/1800, https://github.com/snowplow/snowplow/issues/1801 and https://github.com/snowplow/snowplow/issues/1802.


alexanderdean:
@danisola - this looks great. Thanks so much for putting this together!

1. @danisola - could you move the shredder into `utils.shredder`, just so the association between the Shredder and the TypeHierarchy is more obvious
2. @fblundun - can you take a look at this PR and let us know any feedback!

danisola:
BTW, notice that this change breaks quite a few tests Scala Hadoop Enrich, because they have custom contexts that don't have a schema. An example is [CollectorPayload1LzoSpec](https://github.com/snowplow/snowplow/blob/master/3-enrich/scala-hadoop-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.hadoop/good/CollectorPayload1LzoSpec.scala), which has the custom context `{"page":{"url":"blog"}}`.

Is it just a matter to fix those tests? Or are these non-schemed contexts valid?

alexanderdean:
Hey @danisola - ah yes, I think those schema-less contexts are from before we came up with self-describing JSON. So yep those tests just need to be updated to use any simple schema that is available in Iglu Central. Here is a super-simple schema I just published today:

https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/uri_redirect/jsonschema/1-0-0

fblundun:
Is it worth having a configuration option to disable JSON validation in Scala Common Enrich for the benefit of people who have historical raw data containing schema-less contexts and unstructured events?

alexanderdean:
Interesting idea but JSON validation is such a key part of the platform, I don't see that as workable in the medium term. If we can't validate people's JSONs, very soon we will be able to literally do nothing with their event payloads.

alexanderdean:
I wanted to get more of a handle on the historical raw data as raised by @fblundun:

* [14 May 2013](http://snowplowanalytics.com/blog/2013/05/14/snowplow-0.8.3-released-with-unstructured-events/) we released a JS Tracker which supported unstructured events (name and properties), but no support for processing those unstructured events
* [11 Apr 2014](http://snowplowanalytics.com/blog/2014/04/11/snowplow-0.9.1-released-with-initial-json-support/) we released a Snowplow version which wrote ue_name and ue_properties to atomic.events
* [3 July 2014](http://snowplowanalytics.com/blog/2014/07/03/snowplow-javascript-tracker-2.0.0-released/) we released a JS Tracker which replaced name+properties with self-describing JSON
* [9 July 2014](http://snowplowanalytics.com/blog/2014/07/09/snowplow-0.9.5-released-with-json-validation-shredding/) we released a Snowplow version with automatic shredding into Redshift. This is essentially the version we support today

Two interesting observations:

* We silently dropped ue_na from the Tracker Protocol v2 (not good practice)
* We removed ue_na support from Common Enrich without any kind of migration path - presumably because we knew that a ue_na value wouldn't be a valid Iglu schema URI

Based on this analysis, I think we go ahead with the change as planned. I think if somebody has a large historical archive of schema-less unstructured events which they can no longer re-process, the solution is a manual EMR job to turn those legacy JSONs into self-describing JSONs. If this is very widespread, we could even add a SnowplowAdapter which does that (it would be functionally similar to a MixpanelAdapter or KeenAdapter anyway).

3-enrich/scala-hadoop-shred/src/main/scala/com.snowplowanalytics.snowplow.enrich/hadoop/ShredJob.scala
```diff
@@ -146,7 +146,7 @@ class ShredJob(args : Args) extends Job(args) {
       ShredJob.projectBads(o)
     }
     .mapTo(('line, 'errors) -> 'json) { both: (String, ProcessingMessageNel) =>
-      BadRow(both._1, both._2).toCompactJson
+      new BadRow(both._1, both._2).toCompactJson
```
fblundun: I think it's cleaner to keep this without the `new`.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentManager.scala
```diff
@@ -17,7 +17,10 @@ package common
 package enrichments
 
 // Joda
```
fblundun: This comment has become detached from the DateTime import

fblundun:
I've made comments for all the feedback I have. It's pretty superficial.

3-enrich/scala-hadoop-shred/src/main/scala/com.snowplowanalytics.snowplow.enrich/hadoop/ShredJob.scala
danisola: Without the `new` the compiler interprets that we want to use [BadRow#apply](https://github.com/snowplow/snowplow/blob/feature/r71/3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/outputs/BadRow.scala#L48), which uses Strings instead of ProcessingMessages.

I just added the new because it was the smallest change to make the code compile. Otherwise I could create a new `BadRow#apply` that uses ProcessingMessages or I could `both._2.map(_.toString)`.

danisola:
Should I continue working on this branch for the other tickets? Or do you want me to open new PR that branch from this one?

alexanderdean:
I think this branch is fine for all of 'em!

3-enrich/scala-hadoop-shred/src/main/scala/com.snowplowanalytics.snowplow.enrich/hadoop/ShredJob.scala
fblundun: Fair enough!

danisola:
The test in Scala Hadoop Enrich fail, but before fixing them (there are a few of them...!) I wanted to push https://github.com/snowplow/snowplow/commit/8d809b6d5c746a5cecf412f5564a3cf6ef12efbd to make sure the new fields are in the correct position in `EnrichedEvent`.

danisola:
I've moved the fields to the end of `EnrichedEvent`. Now the tests in Hadoop Enrich pass, but it's just because all the fields after 'document' are not being tested (currency, geolocation, etc)

danisola:
Regarding https://github.com/snowplow/snowplow/issues/1801 and https://github.com/snowplow/snowplow/issues/1802, what are the migration script versions I should work on? Are they `0.5.0` to `0.6.0` for Postgres and `0.6.0` to `0.7.0` for Redshift?

alexanderdean:
@danisola - yep those migration versions sound good. Remember that because the new fields are at the end, the migration scripts can just use `ALTER TABLE ADD COLUMN` statements (would put them all in a single transaction)...

danisola:
I think I've finished with these tickets! Let me know if I've missed or you want me to change anything.

alexanderdean:
Legend! Thanks so much Dani

fblundun:
Thanks @danisola - I'll add your commits into the feature/r71 branch!

alexanderdean:
Exciting!

danisola:
Thank you for the direction! :+1:

Issue #998, opened 2014-09-21
Scala Common Enrich: add cookie extractor enrichment


alexanderdean:
For companies using the Scala Stream Collector on a first-party domain, it would be neat to have an enrichment which could:

* Take a list of cookies you were interested in
* Generate a context for each cookie. Each cookie would have the same structure: name & value

This is neat because companies often have valuable info (e.g. about CMS sessions etc) in their first-party cookies.

This would only work currently for the Scala Stream Collector (only collector which records headers and thus cookies).

kazjote:
I am working on it. These are my propsed json schemas:

https://github.com/snowplow/iglu-central/compare/master...RetentionGrid:cookie_extractor?expand=1

fblundun:
I think the cookie_extractor config schema looks good.

For the cookie_extractor_context - should have one context per cookie rather than all cookies in a single context? I think this would be more elegant in Redshift since we would avoid having a field containing a raw literal JSON array string.

Thoughts @yalisassoon ?

alexanderdean:
Duplicate of #2072

PR #2072, opened 2015-09-26
Scala Common Enrich: add cookie extractor enrichment

kazjote:
Implements #998

### DIFF
--- a/3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/CookieExtractorEnrichment.scala
+++ b/3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/CookieExtractorEnrichment.scala
@@ -1,0 +1,100 @@
+/*
+ * Copyright (c) 2012-2014 Snowplow Analytics Ltd. All rights reserved.
+ *
+ * This program is licensed to you under the Apache License Version 2.0,
+ * and you may not use this file except in compliance with the Apache License Version 2.0.
+ * You may obtain a copy of the Apache License Version 2.0 at http://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * Unless required by applicable law or agreed to in writing,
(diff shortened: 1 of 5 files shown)


3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/CookieExtractorEnrichment.scala
```diff
@@ -0,0 +1,100 @@
...
+          Some(filtered)
+        case _ => None
+      }
+    }.flatten
+
+    cookies.map { cookie =>
+      ( ("schema" -> "iglu:com.snowplowanalytics.snowplow/cookie_extractor_context/jsonschema/1-0-0") ~
+        ("data" ->
+          ("name" -> cookie.getName) ~
+          ("value" -> cookie.getValue)
```
fblundun: For a malformed header like "Cookie: first=a;second", I believe this will give us the JSON

```json
{
  "first": "a",
  "second": null
}
```

which would not pass the schema (since null values are not allowed).

I don't know what the best approach is here - we could fail the whole event, or silently remove the bad headers, or alter the schema to allow nulls. @alexanderdean might have an opinion when he gets back from leave.

It shouldn't matter much since malformed cookie headers ought to be pretty rare.

fblundun:
This looks great to me @kazjote - thanks very much!

kazjote:
Thanks for working on this great project! :-)

alexanderdean:
This looks good! Thanks so much @kazjote. Do we want to support some kind of wildcard "*" option for people who want to capture all cookies?

kazjote:
I was thinking about this. However, I was afraid that in majority cases it would mean importing session cookies into analytic platform. Because of this, I think that '*' filter option could be a trap rather than help.

alexanderdean:
Hi @kazjote - okay cool. I've created a ticket to re-consider this wildcard option at some future point: #2075

alexanderdean:
This has gone into the branch for release, closing.

PR #2142, opened 2015-10-25
StorageLoader: Add skip_missing_jsonpaths option to simply ignore shredded events without a jsonpaths declaration

epantera:
Sometime it could be very useful to do now import all shredded events in database.
This `skip_missing_jsonpaths` option will simply help to ignore shredded events with no jsonpaths file declaration in iglu.

**Why ?**

Self Described Json are a really powerful concept. When you decided to adopt it, it could spread pretty quickly in all yours applications.

Here at Viadeo we have more than a thousand of custom events, in Mobile apps, event-sourced platform, and web front-ends. We have chosen Self Described Json as a standard to describe our events and to fire them with Snowplow Trackers.

But while we are pretty happy with that and happy to keep all those events into our S3 buckets, we do not want to create a Redshift Tables for each custom events.

With this new StorageLoader config option, the iglu jsonpaths file become optional for a custom event. If exists so the custom event will be imported in a dedicated Redshift table. If not, the custom event will simply be archived and not imported in a Redshift table.

### DIFF
--- a/3-enrich/emr-etl-runner/config/config.yml.sample
+++ b/3-enrich/emr-etl-runner/config/config.yml.sample
@@ -55,6 +55,7 @@ enrich:
   continue_on_unexpected_error: false # Set to 'true' (and set :out_errors: above) if you don't want any exceptions thrown from ETL
   output_compression: NONE # Compression only supported with Redshift, set to NONE if you have Postgres targets. Allowed formats: NONE, GZIP
 storage:
+  skip_missing_jsonpaths: false # Set true to prevent to ignore importing an existing shredded event when no jsonpath is defined
   download:
     folder: # Postgres-only config option. Where to store the downloaded files. Leave blank for Redshift
   targets:
--- a/4-storage/storage-loader/lib/snowplow-storage-loader/redshift_loader.rb
+++ b/4-storage/storage-loader/lib/snowplow-storage-loader/redshift_loader.rb
@@ -112,15 +112,17 @@ module Snowplow
 
             jsonpaths_file = st.discover_jsonpaths_file(s3, config[:aws][:s3][:buckets][:jsonpath_assets])
             if jsonpaths_file.nil?
-              raise DatabaseLoadError, "Cannot find JSON Paths file to load #{st.s3_objectpath} into #{st.table}"
-            end
-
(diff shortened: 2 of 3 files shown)


alexanderdean:
Hey @epantera - thanks for this. Your use case makes a lot of sense and I can see how this setting will work for you with your current setup.

The difficulty with using JSON Paths files being present as a determinant of what to load into Redshift is that we are slowly re-architecting towards auto-generation of JSON Paths files (and Redshift DDL) in Iglu, on demand as required. This step will massively simplify the setup and maintenance of a company-wide self-describing JSON dictionary; it will also make the load process less fragile and finally make it possible to drip-feed Redshift robustly from Kinesis.

Your use case of wanting to not load all JSON entities into a given Redshift database is a 100% valid one. But my hunch is that the right long-term approach to this is some kind of whitelist and/or blacklist filters which can be configured against the storage target...

Does that make sense?

epantera:
Hey @alexanderdean, that totally makes sens. I missed that point. 
Agreed, a kind of whitelist and/or blacklist filters seems the solution.

We will be proactive on that point, it's important for us. I'll try to propose you something.
Any rough thoughts are welcome.

alexanderdean:
Hi @epantera - all sounds good.

I don't have any strong thoughts on the filtering implementation yet - but a couple of points:

1. We will be working on the new Redshift/JSON Paths functionality in Iglu in a set of internal ("ghost") releases - you can see the first one here: https://github.com/snowplow/iglu/issues?q=is%3Aopen+is%3Aissue+milestone%3A%22Scala+repo+server+0.2.0-g1%22
2. We are fairly quickly outgrowing embedding our storage target configurations in the config.yml. The plan (not yet written down anywhere) is to move these out into individual JSON configurations, similar to how we configure the enrichments

BenFradet:
StorageLoader has been ported to [RDB Loader](https://github.com/snowplow/snowplow-rdb-loader) so unfortunately the pull request itself doesn't make sense anymore.

@chuwy do you think the requirement (skipping loading of some shredded event types through whitelist / blacklist or some other mechanism) makes sense? If so please log an issue.

Sorry for the lack of follow up and thank you for your contribution, closing.

Issue #456, opened 2014-01-04
Scala Common Enrich: add weather enrichment


alexanderdean:
Resources:

* http://www.simoahava.com/web-development/universal-analytics-weather-custom-dimension/
* http://mattaster.com/weather-universal-analytics/
* http://www.wunderground.com/weather/api/d/pricing.html
* http://www.ncdc.noaa.gov/oa/climate/isd/
* http://openweathermap.org/API

fblundun:
A first step here would be to reboot https://github.com/snowplow/scala-weather to be a Scala client of Open Weather Map with an LRU cache (obviously not necessary if there's an excellent Java client already, but we will need some LRU caching).

christoph-buente:
Or maybe a background worker could fetch the bulk data hourly and all lookups would be local then. http://openweathermap.org/current#bulk

christoph-buente:
Oh, i just saw bulk downloading comes with a price of at least $470/month

PR #2254, opened 2015-12-17
Feature/weather

alexanderdean:


3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/WeatherEnrichment.scala
```diff
@@ -0,0 +1,186 @@
...
+   *
+   * @param latitude enriched event optional latitude (probably null)
+   * @param longitude enriched event optional longitude (probably null)
+   * @param time enriched event optional time (probably null)
+   * @return weather stamp as self-describing JSON object
+   */
+  // It accepts Java Float (JFloat) instead of Scala's because it will throw NullPointerException
+  // on conversion step if `EnrichedEvent` has nulls as geo_latitude or geo_longitude
+  def getWeatherContext(latitude: Option[JFloat], longitude: Option[JFloat], time: Option[DateTime]): Validation[String, JObject] =
+    if (latitude == null || longitude == null || time == null) {
```
alexanderdean: @chuwy how could any of these three values be null? You are testing Option-boxed values...

alexanderdean:
Looking good! Can you change this commit message to be past tense:

     Scala Common Enrich: add weather enrichment (close #456) 

to:

    Scala Common Enrich: add*ed* weather enrichment (close #456)

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/WeatherEnrichment.scala
chuwy: I added this check because ran into NullPointerException on `Float => JFloat` conversion several times.
```scala
import java.lang.{ Float => JFloat }
class A {
  @BeanProperty var latitude: JFloat = _
}
val a = new A
def get(foo: Option[Float]) = 1  // It will rail NPE
get(Option(a.latitude))
```
I don't think there's a big chance above values could be nulls because we control calling code, but overall this check adding safety for me.

alexanderdean: Okay cool - thanks for clarifying...

PR #2422, opened 2016-02-16
Release/r78

fblundun:


fblundun:
Note: when we rebase this PR later, we only want the commits after c23de2e.

PR #2515, opened 2016-03-09
Update Elasticity config for aws session token

13scoobie:
Updated the Elasticity config to support pulling in InstanceProfileCredentials if config is set to iam, else, pull from config for user + access key


13scoobie:
I apologize - was making a change locally to our fork for testing and code review and had selected snowplow as comparison branch.

Thanks!
-dave

alexanderdean:
No worries - looks like it's coming along!

Issue #2524, opened 2016-03-11
Scala Stream Collector: configurable cookie path


christoph-buente:
The snowplow collector does not set an explicit cookie path when dropping the 3rd party network_id cookie. According to wikipedia (https://en.wikipedia.org/wiki/HTTP_cookie) it does not have an effect, as the tracking endpoint is at /i, which will lead implicitly to an path set to "/" in the browser. But the track link endpoint is a nested resource at "/r/tp2" which would lead to set the cookie path to "/r" which is probably not desirable.

So i think the collector should either statically set the cookie path to "/" to make sure it is valid for all collector endpoints or let it be set in the config page. What do you think?

alexanderdean:
Good catch, I think we should set to "/"

gyurist:
It seems like the same issue affects the clojure collector as well. 
Since no explicit path is set the default browser behavior is used which sets the `sp` cookie path to `/com.snowplowanalytics.snowplow`.

alexanderdean:
Bringing forwards (R82 is in test already)

alexanderdean:
Thanks for flagging @gyurist - created #2739 to explore this...

jbeemster:
Closing this ticket in favour of PR

PR #2528, opened 2016-03-12
Scala Stream Collector: configurable cookie path

christoph-buente:
Hardwire the cookie path to / to make sure it also is valid for nested urls like click track.

This solves #2524


alexanderdean:
Thanks! Scheduling

alexanderdean:
Thanks so much @christoph-buente - scheduling!

christoph-buente:
Did this PR ever make it into upstream? I cannot see it in the current master.

alexanderdean:
Bringing forwards (R82 is in test already)

alexanderdean:
Thanks @christoph-buente - this did get somewhat forgotten. Have given it a concrete release now.

Issue #928, opened 2014-07-28
Scala Hadoop Shred: add ability to filter based on event type


alexanderdean:
The idea comes from #927 - basically, have a list of event types that you are interested in shredding, which is passed in as a configuration option.

The idea is that a company with 500 event types may only be interested in shredding a subset into dedicated Redshift tables.

kala725:
@alexanderdean : My basic question was that only, How to define that which event to shred, as a configuration option?

alexanderdean:
Or filter based on app_id - this is driven by the insight that a Snowplow user may have some apps with huge event volumes (e.g. in-game events, ad impressions) that they want to keep out of Redshift, but some apps with manageable event volumes (e.g. ecomm transactions, admin UI usage) that they want to analyze in Redshift.

Given these various use cases, it would probably be best to support arbitrary filtering logic (JavaScript based?).

BenFradet:
dupe of snowplow/snowplow-rdb-loader#59, closing

PR #2537, opened 2016-03-16
Fix for issue #928

kala725:
Blacklisting in the shredding process for some event types can be achieved by this. Just pass the arguments while running the command --skip-shred <table_name_to_be_skipped> and it works. 

Please review this and let me know, if any improvements can be made.

Refs : #928


4-storage/storage-loader/lib/snowplow-storage-loader/redshift_loader.rb
```diff
@@ -133,17 +133,21 @@ def get_shredded_statements(config, target, s3)
           schema = extract_schema(target[:table])
 
           ShreddedType.discover_shredded_types(s3, config[:aws][:s3][:buckets][:shredded][:good], schema).map { |st|
-
-            jsonpaths_file = st.discover_jsonpaths_file(s3, config[:aws][:s3][:buckets][:jsonpath_assets])
-            if jsonpaths_file.nil?
-              raise DatabaseLoadError, "Cannot find JSON Paths file to load #{st.s3_objectpath} into #{st.table}"
+            if config[:skipshred].include?(st.table)
+              #actions to trigger while escaping the table shreding
+		          SqlStatements.new("ANALYZE atomic.events")
```
fblundun: A couple of things here:

* We don't automatically assume that the table is called "atomic.events" - the user can set the name of the table in the configuration file
* It looks like this will run `ANALYZE atomic.events` once per skipped shredded type. We only want to run it once.

I don't think we actually need to create any ANALYZE statements here, because they are already created [here](https://github.com/snowplow/snowplow/blob/b772f6e089bb9e0b3bdbed93a4ab18902c4f1a4a/4-storage/storage-loader/lib/snowplow-storage-loader/redshift_loader.rb#L108). So for skipped types we don't need to add anything to the array returned by get_shredded_statements. One way to do this would be to return `nil` to the array of SqlStatements for each skipped type, then call `.compact` on the resulting array to remove the `nil`s.

fblundun:
Thanks for this @kala725 ! It looks good - I have just made a comment on the code about avoiding unnecessary ANALYZE statements.

kala725:
Thanks Fred. I will make the necessary changes and submit it again. Thanks again for the prompt reply. @fblundun

fblundun:
@kala725 your latest commit looks good to me. Could you sign our [CLA](https://github.com/snowplow/snowplow/wiki/CLA) please?

kala725:
Sure Fred .

alexanderdean:
Hey @kala725 - thanks so much for this! Very exciting contribution to Snowplow.

I have a couple of additional pieces of feedback:

* The command-line arguments like `--skip` are really more about controlling the overall jobflow DAG (i.e. which steps to run), rather than the specifics of how each step should operate (as e.g. which shredded types to include/exclude). *Therefore*, it would be better to include the skip options in the `config.yml` than at the command-line. Especially because you will probably make a decision on which types to load or not load and then re-use that list each run
* If we are not loading a shredded type into a database, there is no point paying the "AWS tax" of processing it through to S3 etc. Therefore over time we will most likely move the type exclusion code out of StorageLoader into Hadoop Shred. We don't need to do this now though - but I've created a ticket to track this: https://github.com/snowplow/snowplow/issues/2540

Here's a suggested format for `config.yml` for further discussion:

```yaml
  targets:
    - name: "Snowplow REDSHIFT Database - us-east-1"
      type: "redshift"
      shredded_types:
        include:
          - com.acme/...
        OR:
        exclude:
          - com.acme/
      ...
```

kala725:
Thanks @alexanderdean for the feedback. I am fine with the first point and will make the changes as suggested, However about second point, my use case was a bit different. I want to store it in atomic.events table as an unstruct event and didn't wanted to create individual tables for some specific event types, So that I can do my analysis in a fast manner (user level targetting in a specific manner, my algo requires it.), also can do my research on some event types (which I want to shred) in more detailed fashion.

alexanderdean:
Hey @kala725 - on point 2, I'm just suggesting moving the equivalent whitelist/blacklist code upstream into Hadoop Shred. You would still see entries in atomic.events for events which were excluded... If I have misunderstood your point, let me know!

kala725:
@alexanderdean , just a question. My colleage  @sambhavsharma contributed to snowplow. Can you tell, which CLA he signed, individual or corporate? 

The pr was https://github.com/snowplow/referer-parser/pull/109

alexanderdean:
Hi @kala725 - yes sure, he signed the **Individual** CLA

kala725:
@alexanderdean : Made the changes suggested by you of taking the excluded entries from config. Updating the CLA.

3-enrich/emr-etl-runner/config/config.yml.sample
```diff
@@ -82,6 +85,9 @@ storage:
       password: ADD HERE
       maxerror: # Not required for Postgres
       comprows: # Not required for Postgres
+      shredded_types:
+        exclude:
+          - ADD HERE  # Leave as it is , even if you don't want to exclude any event type from shredding
```
alexanderdean: We don't yet include shredding for PG, so I would replace with:

```yaml
    exclude: []
```

3-enrich/emr-etl-runner/config/config.yml.sample
```diff
@@ -95,6 +101,9 @@ storage:
       es_nodes_wan_only: false # Set to true if using Amazon Elasticsearch Service
       maxerror: # Not required for Elasticsearch
       comprows: # Not required for Elasticsearch
+      shredded_types:
+        exclude:
+          - ADD HERE # Leave as it is , even if you don't want to exclude any event type from shredding
```
alexanderdean: We only support loading bad rows into ES currently, so I would replace with:

```yaml
    exclude: []
```

4-storage/storage-loader/lib/snowplow-storage-loader/redshift_loader.rb
```diff
@@ -133,18 +133,21 @@ def get_shredded_statements(config, target, s3)
           schema = extract_schema(target[:table])
 
           ShreddedType.discover_shredded_types(s3, config[:aws][:s3][:buckets][:shredded][:good], schema).map { |st|
-
-            jsonpaths_file = st.discover_jsonpaths_file(s3, config[:aws][:s3][:buckets][:jsonpath_assets])
-            if jsonpaths_file.nil?
-              raise DatabaseLoadError, "Cannot find JSON Paths file to load #{st.s3_objectpath} into #{st.table}"
+            if target[:shredded_types][:exclude].include?(st.table)
```
alexanderdean: It looks like we are checking the list of shredded types against the table name, rather than the type name?

alexanderdean:
Thanks! I've left some feedback. As background on my last comment - I'm keen that we white/blacklist explicit JSON types (e.g. `com.acme/shopping_cart/...`), rather than table names.

This is because table names are an implementation detail (may change one day; may have different naming rules per database type), whereas the JSON types are the formal representation of the entities themselves.

kala725:
@alexanderdean : Got it. A small help, Can you tell, how can I get it from predefined functions , if any . I am getting the complete s3_objectpath. Can I get the schema with some already defined function, or I have to write it.

alexanderdean:
@fblundun could you help out @kala725 here? I think it will be about taking the last four sub-folders off of the s3_objectpath...

fblundun:
@kala725 - ShreddedType's initialization logic involves deriving the `vendor` and `name` parts of the schema from the s3_objectpath [here](https://github.com/snowplow/snowplow/blob/b772f6e089bb9e0b3bdbed93a4ab18902c4f1a4a/4-storage/storage-loader/lib/snowplow-storage-loader/shredded_type.rb#L72-L73). So you could extract the vendor and name from each blacklisted schema, and ignore each ShreddedType returned by `discover_shredded_types` which matches both.

I don't think we currently have a Ruby function to extract the name and vendor from a schema string, But a regular expression along the lines of `/([^\/]*)\/([^\/]*)/` should do it. e.g.

```ruby
input = 'com.acme/shopping_cart/...'
match = /([^\/]*)\/([^\/]*)/.match(input)
match[1] == 'com.acme'
match[2] == 'shopping_cart'
```

Does that make sense?

kala725:
Thanks @fblundun @alexanderdean . However I did in a different fashion. Please review my last commit.

kala725:
Hi @fblundun @alexanderdean , Just wanted to know that the code is good to go or requires modification ??

4-storage/storage-loader/lib/snowplow-storage-loader/redshift_loader.rb
```diff
@@ -134,17 +134,26 @@ def get_shredded_statements(config, target, s3)
...
-            jsonpaths_file = st.discover_jsonpaths_file(s3, config[:aws][:s3][:buckets][:jsonpath_assets])
-            if jsonpaths_file.nil?
-              raise DatabaseLoadError, "Cannot find JSON Paths file to load #{st.s3_objectpath} into #{st.table}"
+            # inherited from storage-loader/lib/snowplow-storage-loader/shredded_type.rb @initialize
+            s3_objectpath_parts = /^.*\/(?<vendor>[^\/]+)\/(?<name>[^\/]+)\/(?<format>[^\/]+)\/(?<version_model>[^\/]+)-$/.match(st.s3_objectpath)
+
+            iglu_schema_path = File.join(s3_objectpath_parts[:vendor], "/",  s3_objectpath_parts[:name], "/", s3_objectpath_parts[:format], "/", s3_objectpath_parts[:version_model])
+
+            if target[:shredded_types][:exclude].include?(iglu_schema_path)
+		      nil
```
alexanderdean: Indentation is a little off

alexanderdean:
Thanks @kala725 - @fblundun can you take a look?

fblundun:
It looks good to me. Only one question. At the moment it looks like you have to explicitly provide the `shredded_types.exclude` array even if you don't want to exclude any types. Should it default to an empty array if the setting is not included, for backwards compatibility? Or is that overcomplicating the config?

alexanderdean:
That would be nice...

kala725:
@fblundun @alexanderdean Thanks for the suggestion. Missed the backward compatibility completely. Now added that fix too. Also signed the CLA .

alexanderdean:
CLA signed confirmed! Thanks so much.

kala725:
@alexanderdean : Is there anything needs to be done in this pr, or it is ready to merge.

alexanderdean:
Hey @kala725 - it's functional as it stands, but we may want to revisit the implementation.

The problem with the approach in this PR is that it performs the filtering "late", by excluding files in S3 from being loaded. The issue is that if I exclude `com.acme/checkout`, although the `com_acme_checkout_1` table will **not** be populated, those checkout events will still be present in `atomic.events`. In other words, it's not a comprehensive filtering, just a partial filtering.

Why does this matter? Because if I have an event type which is say 100x the volume of other events and I want to exclude it from loading into Redshift, then this PR as it stands will still load that huge volume into `atomic.events`...

For this reason, we probably want to jump straight to implementing filtering in our Hadoop Shred component, rather than 'late' in the StorageLoader.

Sorry for the late change of view on this... Open to your thoughts.

kala725:
Hey @alexanderdean, Actually my requirements are like this. Suppose  a case where I want filtered data from atomic.events to be indexded in solr/elastic search (Note here, I want filtered events to be indexed as per my requirement), Along with this, I want my analytics engine, to show/create reports for the all the data logged.


So in the above scenerio, which I am facing now (Hope others will reach this problem soon), How we can handle this.

I thought of the above solution, where I can skip entry in the individual table (as not required) for some event types and also store them in atomic.events, which will lead to me a right way to go ahead with (after this , I will go ahead with the solr implementation plan).

 Along with this, I am also going through other possible combinations. Wish to listen your opinions on the same.

alexanderdean:
Thanks for sharing that @kala725 - helpful input!

alexanderdean:
Descheduling this for now

kala725:
@alexanderdean  : This PR is there from a long time. Can you please update me  on the same ?

snowplowcla:
@chuwy can you comment on this PR?

alexanderdean:
@chuwy can you comment on this PR?

chuwy:
I just agree this functionality must be implemented in Shredder, we already ported a ticket for that: https://github.com/snowplow/snowplow-rdb-loader/issues/59. I also agree on a way it should be configured: iglu URIs in `config.yml`, einding up in a common Shredder/Loader [config](https://github.com/snowplow/snowplow-rdb-loader/issues/21).

I think we can close this.

alexanderdean:
Agreed, closing. Thanks again @kala725 for your work here - it's appreciated. Although it hasn't been merged in this form, it will provide a lot of guidance for when we do implement it.

PR #2552, opened 2016-03-21
Friendly error messages

bernardosrulzon:
I just bumped into an EMR problem and it took me a while to figure out what the problem was and which steps I should skip. This PR includes a friendly error message to help with troubleshooting.


alexanderdean:
Hey @bernardosrulzon - thanks for this. We agree that it's quite confusing what resolution you should attempt if the job fails - @ihortom is currently working on a flowchart mapping the job steps to their corresponding data moves, and will start documenting different resolutions for the different cases.

We can't merge your specific suggestion here, because the solution `Try running with --skip staging,enrich` won't be correct in all cases. For example, at Snowplow, the `DirectoryNotEmptyError` is almost always because a prior run has not yet completed, rather than because the corresponding archive step has failed.

However, I think your idea of friendly error messages makes a lot of sense. My suggestion is that once @ihortom has finished his troubleshooting guide, we link to that guide from these error messages. Thoughts?

bernardosrulzon:
Ah, good to see you're working on this! Makes sense - I'll close this PR.

Cheers!

Issue #1184, opened 2014-11-13
Scala Common Enrich: add POST support to IgluAdapter


alexanderdean:
For third-parties that can post an Iglu-compatible self-describing JSON to a Snowplow collector

alexanderdean:
@jbeemster - please work with @chuwy to get this feature into Anton's release branch!

PR #2698, opened 2016-06-02
Adding support for POST JSON payloads in IgluAdapter

miike:
RFC for allowing POST requests to the Iglu webhook endpoint using JSON

### DIFF
--- a/3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/IgluAdapter.scala
+++ b/3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/IgluAdapter.scala
@@ -53,64 +53,42 @@ object IgluAdapter extends Adapter {
   // Create a simple formatter function
   private val IgluFormatter: FormatterFunc = buildFormatter() // For defaults
 
-  /**
-   * Attempts to parse a json string into a JValue
-   * example: {"p":"app"} becomes JObject(List((p,JString(app))))
-   *
-   * @param jsonStr The string we want to parse into a JValue
-   * @return a Validated JValue or a NonEmptyList Failure
-   *         containing a JsonParseException
-   */
-  private[registry] def parseJson(jsonStr: String): Validated[JValue] =
-    try {
-      parse(jsonStr).successNel
-    } catch {
-      case e: JsonParseException => {
-        val exception = JU.stripInstanceEtc(e.toString).orNull
-        s"${VendorName} event failed to parse into JSON: [${exception}]".failNel
-      }
-    }
-
     /**
      *
      * Converts a payload into a list of validated events
      * Expects a valid json - returns a single failure if one is not present
      *
-     * @param body json payload as POST'd by sendgrid
+     * @param body json payload as POST'd by a webhook
      * @param payload the rest of the payload details
      * @return a list of validated events, successes will be the corresponding raw events
      *         failures will contain a non empty list of the reason(s) for the particular event failing
      */
     private def payloadBodyToEvents(body: String, payload: CollectorPayload): List[Validated[RawEvent]] = {
-      try {
 
-        val parsed = parse(body)
-        val queryString = toMap(payload.querystring)
-        val schema = queryString.get("schema") match {
-          case Some(schemaUri) => schemaUri
-        }
-
-        if (parsed.children.isEmpty) {
-          return List(s"$VendorName event failed json sanity check: has no events".failNel)
-        }
-
-        List(RawEvent(
-          api          = payload.api,
-          parameters   = toUnstructEventParams(TrackerVersion,
-                          queryString,
-                          schema,
-                           cleanupJsonEventValues(parsed, None, "timestamp", _ * 1000),
-                           "srv"),
-          contentType  = payload.contentType,
-          source       = payload.source,
-          context      = payload.context
-        ).success)
-      } catch {
-        case e: JsonParseException => {
-          val exception = JU.stripInstanceEtc(e.toString).orNull
-          List(s"$VendorName event failed to parse into JSON: [$exception]".failNel)
-        }
+      val parsed = parseJsonSafe(body) match {
+        case Success(jsonBody) => jsonBody
       }
+
+      val queryString = toMap(payload.querystring)
+      val schema = queryString.get("schema") match {
+        case Some(schemaUri) => schemaUri
+      }
+
+      if (parsed.children.isEmpty) {
+        return List(s"$VendorName event failed json sanity check: has no events".failNel)
+      }
+
+      List(RawEvent(
+        api          = payload.api,
+        parameters   = toUnstructEventParams(TrackerVersion,
+                        queryString,
+                        schema,
+                         parsed,
+                         "srv"),
+        contentType  = payload.contentType,
+        source       = payload.source,
+        context      = payload.context
+      ).success)
     }
 
 
--- a/3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/IgluAdapter.scala
+++ b/3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/IgluAdapter.scala
@@ -55,40 +55,35 @@ object IgluAdapter extends Adapter {
 
     /**
      *
-     * Converts a payload into a list of validated events
+     * Converts a payload into a single validated event
      * Expects a valid json - returns a single failure if one is not present
      *
      * @param body json payload as POST'd by a webhook
      * @param payload the rest of the payload details
-     * @return a list of validated events, successes will be the corresponding raw events
-     *         failures will contain a non empty list of the reason(s) for the particular event failing
+     * @param schemaUri the schemaUri for the event
+     * @param params The query string parameters
+     * @return a single validated event
      */
-    private def payloadBodyToEvents(body: String, payload: CollectorPayload): List[Validated[RawEvent]] = {
-
+    private def payloadBodyToEvent(body: String, payload: CollectorPayload, schemaUri: String, params: Map[String, String]): Validated[RawEvent] = {
       val parsed = parseJsonSafe(body) match {
         case Success(jsonBody) => jsonBody
       }
 
-      val queryString = toMap(payload.querystring)
-      val schema = queryString.get("schema") match {
-        case Some(schemaUri) => schemaUri
-      }
-
       if (parsed.children.isEmpty) {
-        return List(s"$VendorName event failed json sanity check: has no events".failNel)
+        return s"$VendorName event failed json sanity check: has no events".failNel
       }
 
-      List(RawEvent(
+      RawEvent(
         api          = payload.api,
         parameters   = toUnstructEventParams(TrackerVersion,
-                        queryString,
-                        schema,
+                        params,
+                        schemaUri,
                          parsed,
                          "srv"),
         contentType  = payload.contentType,
         source       = payload.source,
         context      = payload.context
-      ).success)
+      ).success
     }
 
 
@@ -128,8 +123,11 @@ object IgluAdapter extends Adapter {
             case(_, None) => "Content type has not been specified".failNel // POST request with no content type
             case (Some(body), Some(contentType)) => payload.contentType match { // POST request with body and content type
               case Some("application/json") =>
-                  val events = payloadBodyToEvents(body, payload)
-                  rawEventsListProcessor(events)
+                  val event = payloadBodyToEvent(body, payload, schemaUri, params)
+                  event match {
+                    case Success(s) => NonEmptyList(s).success
+                    case Failure(f) => f.fail
+                  }
               case _ => "Content type not supported".failNel
             }
           }
--- a/3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/IgluAdapter.scala
+++ b/3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/IgluAdapter.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2014 Snowplow Analytics Ltd. All rights reserved.
+ * Copyright (c) 2014-2016 Snowplow Analytics Ltd. All rights reserved.
  *
  * This program is licensed to you under the Apache License Version 2.0,
  * and you may not use this file except in compliance with the Apache License Version 2.0.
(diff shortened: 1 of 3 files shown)


alexanderdean:
Very cool! Thanks @miike. Assigning to @jbeemster to review. @miike has signed the CLA.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/IgluAdapter.scala
```diff
@@ -47,11 +48,73 @@ object IgluAdapter extends Adapter {
...
   /**
+   * Attempts to parse a json string into a JValue
+   * example: {"p":"app"} becomes JObject(List((p,JString(app))))
+   *
+   * @param jsonStr The string we want to parse into a JValue
+   * @return a Validated JValue or a NonEmptyList Failure
+   *         containing a JsonParseException
+   */
+  private[registry] def parseJson(jsonStr: String): Validated[JValue] =
+    try {
```
jbeemster: Can we move this function to `Adapter.scala` to cut down on code duplication.  Potentially also rename to `parseJsonSafe`.  This function can then be used in the `PingdomAdapter` as well or anywhere else we need to safely parse JSON

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/IgluAdapter.scala
```diff
@@ -47,11 +48,73 @@ object IgluAdapter extends Adapter {
...
+        s"${VendorName} event failed to parse into JSON: [${exception}]".failNel
+      }
+    }
+
+    /**
+     *
+     * Converts a payload into a list of validated events
+     * Expects a valid json - returns a single failure if one is not present
+     *
+     * @param body json payload as POST'd by sendgrid
```
jbeemster: Why is Sendgrid mentioned here?

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/IgluAdapter.scala
```diff
@@ -47,11 +48,73 @@ object IgluAdapter extends Adapter {
...
+     * @param payload the rest of the payload details
+     * @return a list of validated events, successes will be the corresponding raw events
+     *         failures will contain a non empty list of the reason(s) for the particular event failing
+     */
+    private def payloadBodyToEvents(body: String, payload: CollectorPayload): List[Validated[RawEvent]] = {
+      try {
+
+        val parsed = parse(body)
```
jbeemster: In place of wrapping the whole function in a try catch block we can use the `parseJsonSafe` from above.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/IgluAdapter.scala
```diff
@@ -47,11 +48,73 @@ object IgluAdapter extends Adapter {
...
+        val queryString = toMap(payload.querystring)
+        val schema = queryString.get("schema") match {
+          case Some(schemaUri) => schemaUri
+        }
+
+        if (parsed.children.isEmpty) {
```
jbeemster: I am not sure I understand the flow here.  Are we always expecting the body to be a Json Array or can it also be a single Json Object?  As this is a generic adapter it might be nicer to match against the `parsed` value as being either a JArray or JObject and to process accordingly.

If it is a JArray you know that there are multiple events to be processed.

If it is a JObject there is a single event to process.

alexanderdean: I was imaging that the expected payload would always be a (single) self-describing JSON, which would become an unstructured event...

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/IgluAdapter.scala
```diff
@@ -47,11 +48,73 @@ object IgluAdapter extends Adapter {
...
+          return List(s"$VendorName event failed json sanity check: has no events".failNel)
+        }
+
+        List(RawEvent(
+          api          = payload.api,
+          parameters   = toUnstructEventParams(TrackerVersion,
+                          queryString,
+                          schema,
+                           cleanupJsonEventValues(parsed, None, "timestamp", _ * 1000),
```
jbeemster: This line makes sense when we know what fields in the payload actually are to do cleanup - as this Adapter has no knowledge about what Keys are going to be available it feels dangerous to attempt mutation.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/IgluAdapter.scala
```diff
@@ -47,11 +48,73 @@ object IgluAdapter extends Adapter {
...
+                           "srv"),
```
jbeemster: @alexanderdean should we check the QueryString for the `platform` key and set a custom platform if specified?

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/IgluAdapter.scala
```diff
@@ -66,21 +129,32 @@ object IgluAdapter extends Adapter {
...
+                  NonEmptyList(RawEvent(
+                    api          = payload.api,
+                    parameters   = toUnstructEventParams(TrackerVersion, (params - "schema"), schemaUri, IgluFormatter, "app"),
+                    contentType  = payload.contentType,
+                    source       = payload.source,
+                    context      = payload.context
+                    )).success
+            case(_, None) => "Content type has not been specified".failNel // POST request with no content type
+            case (Some(body), Some(contentType)) => payload.contentType match { // POST request with body and content type
+              case Some("application/json") =>
```
jbeemster: A lot of webhooks end up sending a querystring within the payload.body rather than a properly formatted JSON.  To this end we should include a "application/x-www-form-urlencoded" content-type as well.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/IgluAdapter.scala
jbeemster: I was imagining the same but I guess there could be scope for a POST Json Array...

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/IgluAdapter.scala
```diff
@@ -47,11 +48,73 @@ object IgluAdapter extends Adapter {
...
+        }
+
+        if (parsed.children.isEmpty) {
+          return List(s"$VendorName event failed json sanity check: has no events".failNel)
+        }
+
+        List(RawEvent(
+          api          = payload.api,
+          parameters   = toUnstructEventParams(TrackerVersion,
+                          queryString,
```
jbeemster: From what I can see the `schema -> value` pair would still be in the payload here?

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/IgluAdapter.scala
alexanderdean: I can't visualize how an array would be mapped to an event...

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/IgluAdapter.scala
alexanderdean: Happy for that to be in a v2...

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/IgluAdapter.scala
jbeemster: The only situation I can think of is if the Webhook sent groups of the same event as an array.  That does seem unlikely to happen..

jbeemster: I guess I was thrown by it returning a List of RawEvents rather than a single object initially, as there is not actually any processing here to handle multiple events as of yet..

miike:
I think this new commit hopefully addresses and fixes most of the comments made above.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/IgluAdapter.scala
```diff
@@ -47,10 +48,50 @@ object IgluAdapter extends Adapter {
...
+     *
+     * Converts a payload into a list of validated events
+     * Expects a valid json - returns a single failure if one is not present
+     *
+     * @param body json payload as POST'd by a webhook
+     * @param payload the rest of the payload details
+     * @return a list of validated events, successes will be the corresponding raw events
+     *         failures will contain a non empty list of the reason(s) for the particular event failing
+     */
+    private def payloadBodyToEvents(body: String, payload: CollectorPayload): List[Validated[RawEvent]] = {
```
jbeemster: This function is returning a list of Validated RawEvents, however the actual body of the function is only ever going to return a single RawEvent in the List.  This should be changed to return just a `Validated[RawEvent]`.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/IgluAdapter.scala
```diff
@@ -66,21 +107,32 @@ object IgluAdapter extends Adapter {
...
+                    parameters   = toUnstructEventParams(TrackerVersion, (params - "schema"), schemaUri, IgluFormatter, "app"),
+                    contentType  = payload.contentType,
+                    source       = payload.source,
+                    context      = payload.context
+                    )).success
+            case(_, None) => "Content type has not been specified".failNel // POST request with no content type
+            case (Some(body), Some(contentType)) => payload.contentType match { // POST request with body and content type
+              case Some("application/json") =>
+                  val events = payloadBodyToEvents(body, payload)
+                  rawEventsListProcessor(events)
```
jbeemster: The `payloadBodyToEvents` function is doing a lot of extra work again that has already been done such as checking the Schema exists and extracting it.  This block only needs to:

* Parse the JSON body safely.
* Return either a NonEmptyList(String()).failure or a NonEmptyList(RawEvent()).success

There is no need to fetch a list of events as we are only expecting one event and as such no need to to pass it to the rawEventsListProcessor.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/IgluAdapter.scala
```diff
@@ -47,10 +48,45 @@ object IgluAdapter extends Adapter {
...
+     *
+     * @param body json payload as POST'd by a webhook
+     * @param payload the rest of the payload details
+     * @param schemaUri the schemaUri for the event
+     * @param params The query string parameters
+     * @return a single validated event
+     */
+    private def payloadBodyToEvent(body: String, payload: CollectorPayload, schemaUri: String, params: Map[String, String]): Validated[RawEvent] = {
+      val parsed = parseJsonSafe(body) match {
+        case Success(jsonBody) => jsonBody
```
jbeemster: If the parse fails we need to match for Failure as well otherwise this will throw an exception.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/IgluAdapter.scala
```diff
@@ -47,10 +48,45 @@ object IgluAdapter extends Adapter {
...
+        return s"$VendorName event failed json sanity check: has no events".failNel
+      }
+
+      RawEvent(
+        api          = payload.api,
+        parameters   = toUnstructEventParams(TrackerVersion,
+                        params,
+                        schemaUri,
+                         parsed,
+                         "srv"),
```
jbeemster: Could these all be indented the same or even moved back onto the same line?

alexanderdean:
Can we rebase this when ready to a single commit "Scala Common Enrich: add POST support to IgluAdapter (close #1184)"

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/IgluAdapter.scala
```diff
@@ -66,21 +102,35 @@ object IgluAdapter extends Adapter {
...
+                    contentType  = payload.contentType,
+                    source       = payload.source,
+                    context      = payload.context
+                    )).success
+            case(_, None) => "Content type has not been specified".failNel // POST request with no content type
+            case (Some(body), Some(contentType)) => payload.contentType match { // POST request with body and content type
+              case Some("application/json") =>
+                  val event = payloadBodyToEvent(body, payload, schemaUri, params)
+                  event match {
+                    case Success(s) => NonEmptyList(s).success
```
jbeemster: Instead of defining the val and then matching you can directly match onto `payloadBodyToEvent`.  We do need the val afterwards so it is a bit cleaner this way.

jbeemster:
Hey @miike, we are actually going to add some more functionality to this PR so I will be taking it over from here.  Thanks again for the PR!

miike:
@jbeemster No worries. Do you still want me to address the comments above or should I close this PR?

jbeemster:
You can go ahead and close the PR.

Issue #2697, opened 2016-06-02
Scala Stream Collector: add support for cookie bounce


christoph-buente:
## Context

More and more web clients do not support or block 3rd party cookies. The result is, that one user browsing the website results in a series of events, that have the same domain_userid but a lot of different network_userids. For later processing the network_userid is completely useless and could be set to a fixed value like `00000000-0000-4000-A000-000000000000` to adhere to the UUID rules.
## Why

It saves a lot of computation later in the process, because it cuts down the flood of useless network_userids, which will never be seen again.
## How is it done?

Ususally the cookie bounce works like this: The collector takes the request, checks for the presence of the defined cookie name.
- If it is there: use it and process the request.
- If the cookie is missing, the collector issues a redirect to itself with an magic/uniq/special query parameter and the Set-Cookie header. Once the client follows the redirect, the collector checks for the presence of the cookie.
- If it is not presented by the client, it is obvious that 3rd party cookies do not work.
- In this case the network_userid can be set to the defined value and the request could be processed. 
## Implications

This behaviour results in a bit more traffic due to this additional redirect. We observed roughly 30%.

Ideally the value of the network_userid could be specified in the config. If nothing was specified, the collector would work, like it used to be. 

Any suggestions?

alexanderdean:
Discussion is here: http://discourse.snowplowanalytics.com/t/scala-stream-collector-add-support-for-cookie-bounce/306

christoph-buente:
Thx @alexanderdean, but it does not seem to resound with the needs of the rest of the community :)

alexanderdean:
Give it time!

christoph-buente:
⌚ waiting

christoph-buente:
@alexanderdean Can you check our internal pull request?

https://github.com/liveintent-berlin/snowplow/pull/1/files and provide some feedback on it?

PR #2755, opened 2016-06-28
Add support for cookie bounce #2697

christoph-buente:
This is a PR provided by LiveIntent Inc, author is @jankoulaga

We implemented this feature in the least intrusive way. It can be enabled in the config. It is disabled by default. The name of the query param and the fallback UUID can be specified in the config as well.

We provided tests to make sure the current behavior did not change, even though we did a little bit of refactoring. To make the test suite run, we introduced a new dependency to a mocking framework.

Hope you like the functionality. Love to see it be released soon.

FIXES: #2697

### DIFF
--- a/2-collectors/scala-stream-collector/src/main/scala/com.snowplowanalytics.snowplow.collectors/scalastream/ResponseHandler.scala
+++ b/2-collectors/scala-stream-collector/src/main/scala/com.snowplowanalytics.snowplow.collectors/scalastream/ResponseHandler.scala
@@ -22,6 +22,7 @@ import java.nio.charset.StandardCharsets.UTF_8
 import java.util.UUID
 import java.net.URI
 import org.apache.http.client.utils.URLEncodedUtils
+import spray.http.{HttpHeader, HttpResponse, StatusCodes}
 
 // Apache Commons
 import org.apache.commons.codec.binary.Base64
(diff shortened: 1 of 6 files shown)


alexanderdean:
Exciting! Assigning to next open Kinesis release, assigning to @jbeemster to do a first review

christoph-buente:
@jbeemster did you have a chance to look at the code?

christoph-buente:
@alexanderdean is it still scheduled for R86?

alexanderdean:
Hi @christoph-buente - no, this was pushed back in early September... It's a complex feature and we haven't had the bandwidth to do QA on it to date.

christoph-buente:
Hi @alexanderdean, i send some 😘  from 🇩🇪. Would that make it easier to accept our PR and have it released?

christoph-buente:
Hey @alexanderdean or @BenFradet,

R8x is way over. Is there a new milestone, we could attach this PR to?

alexanderdean:
This is already assigned to R92?

BenFradet:
Speaking of which, could you rebase against master or even better against feature/mv-kinesis-es-sink which is the R92 branch?

christoph-buente:
Well, with traffic i mean additional requests. However, the collector checks for cookie and special parameter presence first. So it does not emit extra message to kinesis. From our perspective with ~= 100MM requests a day, it works perfectly fine with a 2-4 collectors, depending on the traffic pattern.

BenFradet:
30% additional requests seems like a pretty steep price to pay.

I'm curious to hear what other people think, @alexanderdean @jbeemster ?

christoph-buente:
Ok, maybe I described it wrong. We see roughly about 30% of the traffic, we cannot put a third party cookie. This leads to one additional request-response cycle for these browsers for every call to the collector. However, the first request coming, that leads to a redirect to the collector is really cheap to handle in terms of computation. And only the second call coming from the redirect leads to a proper analysis of the request and thus a kinesis message.

alexanderdean:
That makes sense. It's certainly a significant cost - but it's an explicit cost that a user will opt in to to avoid downstream complexity/cost. And there isn't an alternative at this (front) end of the pipeline...

christoph-buente:
@alexanderdean You're right, it's a feature that has an "enable" flag, which is off by default. So people need to actively opt in for it. That should not have much impact on anyone, not willing to use it.

2-collectors/scala-stream-collector/src/main/scala/com.snowplowanalytics.snowplow.collectors/scalastream/ResponseHandler.scala
```diff
@@ -96,122 +97,141 @@ class ResponseHandler(config: CollectorConfig, sinks: CollectorSinks)(implicit c
...
+
+    val (httpResponse, badQsResponse) = if (path startsWith "/r/") {
+      // A click redirect
+      try {
+        // TODO: log errors to Kinesis as BadRows
+        val target = URLEncodedUtils.parse(URI.create("?" + queryParams), "UTF-8")
+          .find(_.getName == "u")
+          .map(_.getValue)
+        target match {
+          case Some(t) => HttpResponse(302).withHeaders(`Location`(t) :: headers) -> Nil
```
BenFradet: What happens if we have two location headers because of https://github.com/snowplow/snowplow/pull/2755/files#diff-326d86609fbd1260182e8d32c3e165a0R231 ?

BenFradet:
Reimplemented it in https://github.com/snowplow/snowplow/pull/3274/commits/611242a6e939e1e33c5869e26d9567330c1decfe if you want to have a look, closing

PR #2859, opened 2016-08-16
Scala Common Enrich: leverage the x-forwarded-for field in CloudfrontLoader

danisola:
This PR makes CloudfrontLoader take into account the `x-forwarded-for` field from the [Cloudfront logs](http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/AccessLogs.html#BasicDistributionFileFormat). Related discussion and background info can be found in [Discourse](http://discourse.snowplowanalytics.com/t/x-forwarded-for-in-cloudfront-logs/511).

I've decided to split up the single Regex because, to be correct, the regex should be nesting the newer field groups inside the older field groups. Otherwise it would accept a log line that has the 21 Oct 2013 fields but not the ones added in 12 Sep 2012.

Let me know what you think about the approach.


alexanderdean:
Big thanks @danisola - will take a look!

BenFradet:
integrated into #3824 sorry for the huge delay and thank you for your contribution :+1:

PR #2798, opened 2016-07-07
Call Config.resolve() so environment variables get resolved in hocon

shin-nien:
Allows passing of environment variables so config files can do substitutions

e.g. config file: `cookie { name = "${COOKIE_NAME}"  }`

lets you do: `COOKIE_NAME="example.com" ./snowplow-stream-collector-0.7.0 --config myconfig.conf`

PR #2861, opened 2016-08-18
Make Scala collector resolve variables in config file.

morgante:
This pull request explicitly tells the Scala collector to resolve any variables in the config variable. This enables users to do things like referencing environment variables from the config, which is useful for things like hiding secrets or enabling multiple deployments of the same config.


alexanderdean:
Thanks @morgante! Unfortunately have to close as a duplicate of #2798. Appreciate the contribution though.

PR #2924, opened 2016-11-05
Faster deduplication queries

bernardosrulzon:
Following @bogaert [suggestion on Discourse](http://discourse.snowplowanalytics.com/t/making-sql-data-models-incremental-to-improve-performance-tutorial), I was able to reduce the deduplication runtime by 90% restricting table scans for the previous 4 weeks. 99.9% of our duplicates occur within that time frame.

### DIFF
--- a/5-data-modeling/sql-runner/redshift/sql/deduplicate/01-events.sql
+++ b/5-data-modeling/sql-runner/redshift/sql/deduplicate/01-events.sql
@@ -35,7 +35,22 @@ DROP TABLE IF EXISTS duplicates.tmp_events_id_remaining;
 CREATE TABLE duplicates.tmp_events_id
   DISTKEY (event_id)
   SORTKEY (event_id)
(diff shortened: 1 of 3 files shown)


bogaert:
Hi @bernardosrulzon - thanks for the contribution!

This is indeed a good way to speed up the queries once you have deduplicated the tables a first time. To do so, you'll still need to run the queries without any restrictions on the timestamp. Once that is done, almost all duplicates will indeed be in the last month or so.

I'd be happy to merge these change, but I'd switch the default to having the restrictions commented out - what do you think? If you agree, could you commit these changes to the `feature/web-models` branch? That way it will go out with the next release. The folders in that branch have been restructured a little, but the deduplication queries remain the same.

bernardosrulzon:
Awesome - that makes perfect sense.

BTW - Are you guys working on a new web-incremental model? I've built most of the queries, would be happy to open another PR if you're not working on it.

bogaert:
Hi @bernardosrulzon.

We have built a new web model that does a lot more than the old model (e.g. it calculates scroll depth and page performance). I'll commit it to the `feature/web-models` repo later today. An incremental version will probably follow at some point, but we haven't started on that one yet.

bernardosrulzon:
Closed in favour of #2926

PR #2926, opened 2016-11-07
Faster deduplication queries

bernardosrulzon:
Following @bogaert [suggestion on Discourse](http://discourse.snowplowanalytics.com/t/making-sql-data-models-incremental-to-improve-performance-tutorial/319), I was able to reduce the deduplication runtime by 90% restricting table scans for the previous 4 weeks. 99.9% of our duplicates occur within that time frame.

The restrictions are commented out, as it only makes sense after a user has run a full deduplication once.


bernardosrulzon:
Hey @bogaert - I've commited on your `feature/web-models`. Git newbie here, please check if that worked as you expected.

bogaert:
Thanks @bernardosrulzon. I'll review the commits and merge the PR.

PR #2981, opened 2016-12-02
Release/dedupe

chuwy:


chuwy:
@alexanderdean about overriding existing jars: to overwrite existing asset **all** following conditions should be truth:

* Wrong (previous) version in `Build.scala` (very likely to happen)
* Same wrong version in `TRAVIS_TAG` (unlikely, but still possible)
* `override: true` [in `release.yml`](https://github.com/snowplow/snowplow/blob/release/dedupe/.travis/release_hadoop_shred.yml#L22)

All three these bullets make me pretty confident about safety of published instances.

alexanderdean:
Thanks @chuwy - that gives me a lot of confidence too...

alexanderdean:
Switching over to the release PR...

Issue #3035, opened 2017-01-05
EmrEtlRunner: replace hadoop_shred in config.yml.sample with rdb_shredder


alexanderdean:

BenFradet:
Should we move the config to the storage part as well, instead of enrich?

alexanderdean:
Hmm, good question - I think we should...

alexanderdean:
We're calling the component RDB Shredder, so the config file property should be `rdb_shredder` - updated the ticket title...

BenFradet:
Should we rename the tags as well: https://github.com/snowplow/snowplow/pull/3203/files#diff-354f30a63fb0907d4ad57269548329e3R75 ?

BenFradet:
same goes for https://github.com/snowplow/snowplow/blob/78676a1e2497f0169260986f606b89535a08dff5/.travis/release_relational_database_shredder.yml

alexanderdean:
Let's do it, please rename the tags, commits, tickets etc...

Issue #3033, opened 2017-01-05
EmrEtlRunner: update to run RDB Shredder instead of Scala Hadoop Shred


alexanderdean:

alexanderdean:
Note that this release of EmrEtlRunner should still be capable of running the older versions of Hadoop Shred if desired - it should be handled by a version switch (see existing code for a couple of similar patterns).

Issue #641, opened 2014-04-16
EmrEtlRunner: add ability to run Spark jobs


alexanderdean:

alexanderdean:
Moving this into @BenFradet's milestone...

BenFradet:
I'm wondering if it'd make sense to have a `SparkStep` directly in Elasticity

alexanderdean:
That sounds good to me @BenFradet !

alexanderdean:
You could migrate over the ScaldingStep too...

PR #3048, opened 2017-01-13
Update EmrEtlRunner to run Spark jobs

BenFradet:
- [x] bump to 0.25.0
- [x] scalding step
- [x] test scalding step
- [x] spark step
- [x] ability to run both scalding and spark shreds
- [x] change the config
- [x] test spark step
- [x] bump elasticity
- [ ] update the config in the doc
- [ ] ~~expose spark-specific config params?~~


BenFradet:
I'll treat #3039, #3035, #3033 and #641 in this one

BenFradet:
- [x] bump to 0.25.0
- [x] scalding step
- [x] test scalding step
- [x] spark step
- [ ] test spark step
- [ ] ability to run both scalding and spark shreds
- [ ] bump elasticity
- [ ] change the config

BenFradet:
Welp, I found a nasty compatibility problem between the version of jackson we use compared to Spark's. Hoping both are binary compatible.

alexanderdean:
Welp!

BenFradet:
Apparently, there is code in the iglu-scala-client depending on features of json4s 3.2.11 which are not part of 3.2.10.

chuwy:
@BenFradet is it a `render` signature? I remember there was lot of problems with binary-compatibility in these versions because of `render` function.

Also what dependency uses 3.2.10?

chuwy:
https://github.com/json4s/json4s/issues/212?

BenFradet:
Spark 1.6.1 (the version we have to use in scala 2.10) uses 3.2.10 and since it provides its dependencies itself when launching the job, we have to go back to 3.2.10.

BenFradet:
Currently, the problem comes from `fromJsonNode` added in 3.2.11.

chuwy:
Currently most of our libraries use json4s 3.2.11, including Scala Common Enrich (which uses `fromJsonNode` as well). So, that can be a problem.

chuwy:
@BenFradet short list of our core libs using Json4s 3.2.11:
1.1 iglu client (directly)
1.2 scala common enrich (direcly and though iglu client)
1.3 scala hadoop enrich (though scala common enrich and iglu client)
2.1 scala tracker (directly)
2.2 stream enrich (though scala tracker and scala common enrich)
3.1 scala hadoop shred (directly, though scala common enrich and iglu client) (not so important though, as we're rewritting it)

BenFradet:
All that for those two one-liners :sob:

BenFradet:
@alexanderdean 

@chuwy and I came up with the following strategy:

- downgrade to json4s 3.2.10 all the libraries mentioned above
- move everything back to 3.2.11 when we move to scala 2.11 as later versions of Spark use 3.2.11 but are only built against scala 2.11 in EMR

chuwy:
@BenFradet's another option was to force transition to 2.11. Comparing how many libs we can **downgrade** - this is sane alternative.

BenFradet:
Yup, I'm personally leaning towards moving everything to scala 2.11 as it'd simplify things a fair bit.

alexanderdean:
I vote for the move forwards to Scala 2.11 - it ticks more boxes, even if it takes longer...

BenFradet:
Ok, should we integrate the move to my milestone?

chuwy:
@BenFradet I have some ongoing work on it. I created a milestone to track progress https://github.com/snowplow/snowplow/milestone/139

BenFradet:
Ok, I'll get on it then.

BenFradet:
Should we kill two birds with one stone and move to sbt 0.13.13 also?

alexanderdean:
@BenFradet - to me the move to Scala 2.11 is a dependency on your milestone. So I would do it in a separate branch (so we have flexibility on releasing as 1 or 2 releases), but yes crack on with it please!

BenFradet:
Created both rslifka/elasticity#133 and rslifka/elasticity#134

3-enrich/emr-etl-runner/config/config.yml.sample
```diff
@@ -38,6 +38,7 @@ aws:
       lingual:              # Optional. To launch on cluster, provide version, "1.1", keep quotes. Leave empty otherwise.
     # Adjust your Hadoop cluster below
     jobflow:
+      job_name: Snowplow ETL # Give your job a name
```
BenFradet: Moved the job name here since it's the same job for both enrich and shred.

3-enrich/emr-etl-runner/config/config.yml.sample
```diff
@@ -49,14 +50,14 @@ aws:
...
     hadoop_enrich: 1.8.0 # Version of the Hadoop Enrichment process
-    hadoop_shred: 0.10.0 # Version of the Hadoop Shredding process
-    hadoop_elasticsearch: 0.1.0 # Version of the Hadoop to Elasticsearch copying process
-  continue_on_unexpected_error: false # Set to 'true' (and set :out_errors: above) if you don't want any exceptions thrown from ETL
+  continue_on_unexpected_error: false # Set to 'true' you don't want any exceptions thrown from ETL
   output_compression: NONE # Compression only supported with Redshift, set to NONE if you have Postgres targets. Allowed formats: NONE, GZIP
 storage:
+  versions:
+    relational_database_shredder: 0.12.0 # Version of the Spark Shredding process
+    hadoop_elasticsearch: 0.1.0          # Version of the Hadoop to Elasticsearch copying process
```
BenFradet: Moved hadoop es here as well since it's in storage

BenFradet:
Made a few tests:

- Spark enrich + rdbs :heavy_check_mark: 
- Hadoop enrich + rdbs :heavy_check_mark:
- Spark enrich + Hadoop shred :heavy_multiplication_x: 

Spark enrich + Hadoop shred doesn't fail but it seems the shred job just copy the output of Spark enrich and that's it, still investigating.

rgabo:
If I'm not mistaken, `spark-enrich` will correctly pick up logfiles in nested hierarchies, i.e. if `s3-dist-cp` copies and groups logfiles into HDFS, but keeps the datestamp directories from an archive folder, the Spark job will still work. This has not worked in the past with `hadoop-enrich`, right @alexanderdean? Would be amazing if it did now.

BenFradet:
@rgabo Spark will indeed recursively pick up everything if the paths ends with `/*`

alexanderdean:
It hasn't worked in the past, correct @rgabo. Exciting that this pain finally goes away!

BenFradet:
Spark enrich + Hadoop shred works!

I didn't see any shredded types the first time around that's why I reported it as not working but it must have been because they were too slow to appear, it just happened again.

alexanderdean:
That's great @BenFradet !

pkallos:
:+1:

BenFradet:
closing in favor of #3203

PR #3077, opened 2017-01-25
Spark Enrich

BenFradet:
- [x] rename #3064 
- [x] package #3076 
- [x] 1.9.0 #3072 
- [x] sbt 0.13.13 #3065
- [x] assembly #3078
- [x] port #3067
  - [x] config
  - [x] job
  - [x] test
  - [x] doc
  - [x] rm scalding job
  - [x] rm useless deps
- [x] ci #3069 
- [x] FileUtils
- [x] 2.11 #3070 
- [x] update deps ~~+ rm BeforeAfterAll~~
- [x] rework deploy scripts
- [x] rm empty rdd detection
- [x] investigate symlinks in the distributed cache


BenFradet:
- [x] rename #3064 
- [ ] package #3076 
- [ ] 1.9.0 #3072 
- [ ] sbt 0.13.13 #3065
- [ ] spark #3067
- [ ] 2.11 #3070 
- [ ] ci #3069

BenFradet:
I'm done with the port, except for a few integration tests dealing with thrift, still looking into that.

BenFradet:
All tests are working :+1:

alexanderdean:
Woohoo! What's next @BenFradet - do you need code review?

BenFradet:
I'll add ci/cd and update to 2.11 with a local sce tomorrow but the code shouldn't change much so a review would be much appreciated :+1:

alexanderdean:
Awesome - @chuwy can you do a review for @BenFradet ?

chuwy:
Yep!

3-enrich/scala-hadoop-enrich/project/BuildSettings.scala
```diff
@@ -1,85 +0,0 @@
...
-  import AssemblyKeys._
-  lazy val sbtAssemblySettings = assemblySettings ++ Seq(
-
-    // Simpler jar name
-    jarName in assembly := {
-      name.value + "-" + version.value + ".jar"
-    },
-
-    // Drop these jars
-    excludedJars in assembly <<= (fullClasspath in assembly) map { cp =>
```
chuwy: Did we check that all excludings are unnecessary now?

3-enrich/spark-enrich/project/Dependencies.scala
```diff
@@ -15,27 +15,23 @@ import sbt._
...
   val resolutionRepos = Seq(
-    // Required for our json4s snapshot
-    "Sonatype Snapshots" at "https://oss.sonatype.org/content/repositories/snapshots/",
-    // For some misc Scalding and Twitter libs
-    "Concurrent Maven Repo" at "http://conjars.org/repo",
-    "Clojars Maven Repo" at "http://clojars.org/repo",
-    // For Twitter libs
-    "Twitter maven libs" at "http://maven.twttr.com/",
     // For Snowplow libs
     "Snowplow Analytics Maven repo" at "http://maven.snplow.com/releases/",
```
chuwy: What libraries require these repos? I think we need to move all our dependencies to Maven (and include CI/CD) prior to 2.11 crosspublishing?

3-enrich/spark-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.spark/EnrichJob.scala
```diff
@@ -0,0 +1,173 @@
...
+    // Install MaxMind file(s) if we have them
+    enrichConfig.filesToCache.foreach { case (uri, filename) => sc.addFile(uri.toString) }
+
+    val inFormat = enrichConfig.inFormat
+    val igluConfig = enrichConfig.igluConfig
+
+    val input = getInputRDD(inFormat, enrichConfig.inFolder)
+
+    val common = input
+      .map { line =>
```
chuwy: Can we refactor contents of this tuple into separate values like this:

```scala
      .map { line =>
        val registry = RegistrySingleton.get(igluConfig, enrichConfig.enrichments, enrichConfig.local)
        val loader = LoaderSingleton.get(inFormat).asInstanceOf[Loader[Any]].toCollectorPayload(line)
        val event = EtlPipeline.processEvents(registry, etlVersion, enrichConfig.etlTstamp, loader)(ResolverSingleton.get(igluConfig))
        (line, event)
      }
```

I think it would be more readable even to factor it out as separate function. Basically this small lambda is the heart of Snowplow.

Also, I see that `registry` and
`Loader[Any]` are singletons created once per worker, but is it really necessary to instantiate them straight inside lambda?

3-enrich/spark-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.spark/EnrichJob.scala
```diff
@@ -0,0 +1,173 @@
...
+          case bytes: Array[Byte] => new String(Base64.encodeBase64(bytes), "UTF-8")
+          case other => other.toString
+        }
+        errors.map(BadRow(originalLine, _).toCompactJson)
+      }
+    if (!bad.isEmpty) bad.saveAsTextFile(enrichConfig.badFolder)
+
+    // Handling of properly-formed rows
+    val good = common
+      .flatMap { case (line, enriched) => projectGoods(enriched) }
```
chuwy: `line` is not used, can be substituted with `_`

3-enrich/spark-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.spark/EnrichJob.scala
```diff
@@ -0,0 +1,173 @@
...
+    if (!good.isEmpty)
```
chuwy: newlines before `if`

3-enrich/spark-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.spark/EnrichJob.scala
```diff
@@ -0,0 +1,173 @@
...
+   * @param inFormat Collector format in which the data is coming in
+   * @return A RDD containing strings or byte arrays
+   */
+  private def getInputRDD(inFormat: String, path: String): RDD[_] = {
+    inFormat match {
+      case "thrift" =>
+        import com.twitter.elephantbird.mapreduce.input.MultiInputFormat
+        MultiInputFormat.setClassConf(classOf[Array[Byte]], hadoopConfig)
+        sc.newAPIHadoopFile[
+          org.apache.hadoop.io.LongWritable,
```
chuwy: Why not to import?

3-enrich/spark-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.spark/EnrichJob.scala
```diff
@@ -0,0 +1,173 @@
...
+
+  /**
+   * Read the input files.
+   * @param inFormat Collector format in which the data is coming in
+   * @return A RDD containing strings or byte arrays
+   */
+  private def getInputRDD(inFormat: String, path: String): RDD[_] = {
+    inFormat match {
+      case "thrift" =>
+        import com.twitter.elephantbird.mapreduce.input.MultiInputFormat
```
chuwy: Why `MultiInputFormat` is imported here?

3-enrich/spark-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.spark/EnrichJob.scala
```diff
@@ -0,0 +1,173 @@
...
+}
+
+/**
+ * The Snowplow Enrich job, written in Spark.
+ * @param spark Spark session used throughout the job
+ * @param args Command line arguments for the enrich job
+ */
+class EnrichJob(@transient val spark: SparkSession, args: Array[String]) extends Serializable {
+  @transient private val sc: SparkContext = spark.sparkContext
+  import spark.implicits._
```
chuwy: It seems nothing from `implicits` is used here.

3-enrich/spark-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.spark/EnrichJob.scala
```diff
@@ -0,0 +1,173 @@
...
+ */
+package com.snowplowanalytics.snowplow.enrich
+package spark
+
+// Apache commons
+import org.apache.commons.codec.binary.Base64
+
+// Scalaz
+import scalaz._
+import Scalaz._
```
chuwy: I think we can omit `Scalaz._` import.

3-enrich/spark-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.spark/EnrichJob.scala
```diff
@@ -0,0 +1,173 @@
...
+    val bad = common
+      .map { case (line, enriched) => (line, projectBads(enriched)) }
+      .flatMap { case (line, errors) =>
+        val originalLine = line match {
+          case bytes: Array[Byte] => new String(Base64.encodeBase64(bytes), "UTF-8")
+          case other => other.toString
+        }
+        errors.map(BadRow(originalLine, _).toCompactJson)
+      }
+    if (!bad.isEmpty) bad.saveAsTextFile(enrichConfig.badFolder)
```
chuwy: It would be safer to enclose procedural `if`s into curly braces.

3-enrich/spark-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.spark/EnrichJobConfig.scala
```diff
@@ -0,0 +1,159 @@
...
+      .text("Output folder where the enriched events will be stored")
+    opt[String]("bad-folder").required().valueName("<bad folder>")
+      .action((f, c) => c.copy(badFolder = f))
+      .text("Output folder where the malformed events will be stored")
+    opt[String]("enrichments").required().valueName("<enrichments>")
+      .action((e, c) => c.copy(enrichments = e))
+      .text("JSONs describing the enrichments")
+    opt[String]("iglu-config").required().valueName("<iglu config>")
+      .action((i, c) => c.copy(igluConfig = i))
+      .text("Iglu configuration")
```
chuwy: Iglu *resolver* configuration.

3-enrich/spark-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.spark/EnrichJobConfig.scala
```diff
@@ -0,0 +1,159 @@
...
+  def inFolder: String
+  def inFormat: String
+  def outFolder: String
+  def badFolder: String
+  def enrichments: String
+  def igluConfig: String
+  def local: Boolean
+}
+
+private case class RawEnrichJobConfig(
```
chuwy: I like this `private` approach!

3-enrich/spark-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.spark/EnrichJobConfig.scala
```diff
@@ -0,0 +1,159 @@
...
+      .text("Output folder where the malformed events will be stored")
+    opt[String]("enrichments").required().valueName("<enrichments>")
+      .action((e, c) => c.copy(enrichments = e))
+      .text("JSONs describing the enrichments")
+    opt[String]("iglu-config").required().valueName("<iglu config>")
+      .action((i, c) => c.copy(igluConfig = i))
+      .text("Iglu configuration")
+    opt[Long]("etl-timestamp").required().valueName("<ETL timestamp>")
+      .action((t, c) => c.copy(etlTstamp = t))
+      .text("Timestamp at which the job was launched, in Epoch time")
```
chuwy: Milliseconds?

3-enrich/spark-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.spark/EnrichJobConfig.scala
```diff
@@ -0,0 +1,159 @@
...
+      .text("The format in which the collector is saving data")
+    opt[String]("output-folder").required().valueName("<output folder>")
+      .action((f, c) => c.copy(outFolder = f))
+      .text("Output folder where the enriched events will be stored")
+    opt[String]("bad-folder").required().valueName("<bad folder>")
+      .action((f, c) => c.copy(badFolder = f))
+      .text("Output folder where the malformed events will be stored")
+    opt[String]("enrichments").required().valueName("<enrichments>")
+      .action((e, c) => c.copy(enrichments = e))
+      .text("JSONs describing the enrichments")
```
chuwy: I'd rephrase it to reflect this is path to local dir.

3-enrich/spark-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.spark/singleton.scala
```diff
@@ -0,0 +1,138 @@
...
+package snowplow.enrich
+package spark
+
+// Jackson
+import com.fasterxml.jackson.databind.JsonNode
+
+// Snowplow
+import common.{FatalEtlError, ValidatedMessage, ValidatedNelMessage}
+import common.utils.{ConversionUtils, JsonUtils}
+import iglu.client.Resolver
```
chuwy: Separate project.

3-enrich/spark-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.spark/singleton.scala
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    /**
+     * Build an Iglu resolver from a JSON.
+     * @param json JSON representing the Iglu resolver
+     * @return A Resolver or one or more error messages boxed in a Scalaz ValidationNel
+     */
+    private[spark] def getIgluResolver(json: String): ValidatedNelMessage[Resolver] =
+      for {
+        node <- (base64ToJsonNode(json, "iglu").toValidationNel: ValidatedNelMessage[JsonNode])
```
chuwy: This can be inferred.

3-enrich/spark-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.spark/singleton.scala
```diff
@@ -0,0 +1,138 @@
...
+        reso <- Resolver.parse(node)
+      } yield reso
+  }
+
+  /** Singleton for EnrichmentRegistry. */
+  object RegistrySingleton {
+    import common.enrichments.EnrichmentRegistry
```
chuwy: It's okay to put this import to top (and `fromJsonNode` below)

3-enrich/spark-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.spark/singleton.scala
```diff
@@ -0,0 +1,138 @@
...
+    @volatile private var enrichments: String = _
+    /**
+     * Retrieve or build an instance of EnrichmentRegistry.
+     * @param igluConfig JSON representing the Iglu configuration
+     * @param enrichments JSON representing the enrichments that need performing
+     * @param local Whether to build a registry from local data
+     */
+    def get(igluConfig: String, enrichments: String, local: Boolean): EnrichmentRegistry = {
+      if (instance == null || this.enrichments != enrichments) {
+        synchronized {
```
chuwy: Just out of curiosity - is `synchronized` really necessary? I mean, are these objects really accessed in multi-threading env?

3-enrich/scala-hadoop-enrich/project/BuildSettings.scala
BenFradet: They're not pulled in so yes

3-enrich/spark-enrich/project/Dependencies.scala
BenFradet: AFAIK, sce, collector-payload-1 and snowplow-thrift-raw-event are not published on central.

3-enrich/spark-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.spark/EnrichJob.scala
BenFradet: It brings toDF below

3-enrich/spark-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.spark/EnrichJob.scala
BenFradet: The only way to have them once per worker is to call the singleton inside the lambda.

3-enrich/spark-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.spark/EnrichJob.scala
BenFradet: Because it's specific to this particular format

3-enrich/spark-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.spark/EnrichJob.scala
BenFradet: Because it's used only once and here

3-enrich/spark-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.spark/singleton.scala
BenFradet: All executors are multithreaded so yes

3-enrich/spark-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.spark/EnrichJob.scala
chuwy: It seems `toDF` is part of `DataSet` class. And it also compiles without this import.

3-enrich/spark-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.spark/EnrichJob.scala
chuwy: Well, maybe this is matter of taste, but I'd still import things that are used only once at the top of file.

3-enrich/spark-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.spark/EnrichJob.scala
```diff
@@ -0,0 +1,173 @@
...
+    e => throw FatalEtlError(e.map(_.toString)),
+    identity
+  )
+
+  def run(): Unit = {
+    import EnrichJob._
+    import singleton._
+
+    // Install MaxMind file(s) if we have them
+    enrichConfig.filesToCache.foreach { case (uri, filename) => sc.addFile(uri.toString) }
```
BenFradet: I'm wondering what was the intent of this in the original job since it wasn't used after caching the files

alexanderdean: It should be used inside the enrichment - it's a bit gnarly, but the enrichment should use the Hadoop cache paths that are populated here...

BenFradet: Welp, I guess I'll need to revert back to what was done before since this works a bit differently: it distributes those files on each node and then those files can be accessed inside Spark but I'm assuming the enrichments use HDFS' DistributedCache to access them?

alexanderdean: I'm not wedded to Hadoop's DistCache... One worry I have about continuing to use it is whether it's compatible with running the job on spot instances - if in the future we want to exclusively use spot / not use HDFS, then we will have to migrate off DistCache anyway?

BenFradet: We might yes, however I'm not familiar with how those particular enrichments leverage Hadoop's DistCache and what migrating away from it would entail.

alexanderdean: It's just (so far) the MaxMind IP Lookups enrichment, and it just uses DistCache for storing and retrieving a 20Mb file as its IP address database. I think we can probably safely stick to your new approach...

BenFradet: My concern is how would the ip lookup enrichment find where the files are stored?

alexanderdean: You'd have to do what I do here, which is come up with some standard "global-variable-style" file paths, and then the init code knows where to write to, and the enrichment knows where to read from, without the two components being tightly coupled...

3-enrich/spark-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.spark/EnrichJob.scala
BenFradet: yup, you're right. I'm too used to always having it around

3-enrich/spark-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.spark/EnrichJob.scala
BenFradet: Good idea to factor that out inside its own function, will do.

3-enrich/spark-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.spark/EnrichJob.scala
BenFradet: Welp the problem with my approach I don't have any control over where the files will be stored on each node (/tmp most likely).

alexanderdean: Ah! Yes that sounds problematic

BenFradet:
@chuwy thanks for your review, I think I've taken care of everything

rgabo:
@BenFradet @chuwy @alex this is amazing work, thank you.

We are in the process of reprocessing a lot of archived Snowplow logs (both `cloudfront` and `clj-tomcat`) and I was wondering if you had an ETA for Scala 2.11?

We have in the past written custom pyspark code to parse the logfiles as we wanted to avoid rerunning `hadoop-enrich` and did not need a lot of what enrich produces. We'd much rather rely on `spark-enrich` for the reprocessing, but have a few time sensitive projects in progress.

The moment `spark-enrich` compiles on Scala 2.11, we will start testing it locally and on EMR 5.3.0 and let you know about any issues we face. Data we're planning to process contains everything from JS Tracker 0.9 to 2.x as well as the Java Tracker collected using both CloudFront and Clojure Collectors :)

Thanks for the continuous amazing work that you're doing!

BenFradet:
Hey @rgabo, thanks for the kind words.
I'm in the process of migrating to 2.11, it should land in this PR today or on Monday :+1: .

.travis/deploy_spark_enrich.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
alexanderdean: A little OT but is there a reason why we have all these individual `deploy_${foo}.sh` scripts, when we could just have a singular `deploy.sh ${foo}` approach?

.travis/is_spark_enrich_release_tag.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
alexanderdean: As above, OT but is there a reason why we have all these individual is_${foo}_release_tag.sh scripts, when we could just have a singular is_release_tag.sh ${foo} approach?

.travis/deploy_spark_enrich.sh
BenFradet: I don't know I've just adapted what was already there.

BenFradet: Should we incorporate those changes to this PR?

alexanderdean: I think yes please - it's as good a release as any to rationalise the approach. Thanks @BenFradet

alexanderdean: Will need (one?) new ticket(s)

BenFradet: Ok, I'll work on that on Monday.

BenFradet:
@alexanderdean reworked the travis scripts if you want to have a look :+1:

rgabo:
@BenFradet jumping on the Scala 2.11 support and testing with Spark 2.1. I am going to submit the app without `EmrEtlRunner` for now, but is the current version of `EmrEtlRunner` compatible as far as the job submission goes?

BenFradet:
Nope, we're also modifying EmrEltRunner so it can run Spark jobs in #3048

rgabo:
Looks like #3048 is also coming along nicely. I will lego together what I need and let you know if I bump into any issues. Thanks a ton!

BenFradet:
No problem, I'll also be doing approximately the same tests once I'm done with EmrEtlRunner :+1:

.travis.yml
```diff
@@ -38,26 +38,26 @@ before_deploy:
...
-      condition: '"$(.travis/is_hadoop_shred_release_tag.sh $TRAVIS_TAG)" == "" && $? == 0'
+      condition: '"$(.travis/is_release_tag.sh hadoop_shred $TRAVIS_TAG)" == "" && $? == 0'
   - provider: script
-    script: ./.travis/deploy_hadoop_enrich.sh $TRAVIS_TAG
+    script: ./.travis/deploy.sh spark_enrich $TRAVIS_TAG
     skip_cleanup: true
     on:
       tags: true
-      condition: '"$(.travis/is_hadoop_event_recovery_release_tag.sh $TRAVIS_TAG)" == "" && $? == 0'
+      condition: '"$(.travis/is_release_tag.sh hadoop_event_recory $TRAVIS_TAG)" == "" && $? == 0'
```
BenFradet: Don't really know what this is used for, would love some guidance.

rgabo:
snowplow-common-enrich 0.25.0 does not seem to be published for 2.11 just yet:

```
error] (*:update) sbt.ResolveException: unresolved dependency: com.snowplowanalytics#snowplow-common-enrich_2.11;0.25.0: not found
```

.travis.yml
alexanderdean: Typo, should be recovery not recory

BenFradet:
Yup, I'm currently publishing it locally for the tests.

alexanderdean:
Spotted a typo, otherwise the CI/CD bash refactor looks good @BenFradet

BenFradet:
@alexanderdean Can you share what the hadoop event recovery entry is for?

alexanderdean:
It's for building and deploying this artifact:

https://github.com/snowplow/snowplow/tree/master/3-enrich/hadoop-event-recovery

This artifact is not run as part of the regular Snowplow run, so the changes made in this PR shouldn't affect it...

.travis.yml
```diff
@@ -38,26 +38,26 @@ before_deploy:
...
   - provider: script
-    script: ./.travis/deploy_hadoop_shred.sh $TRAVIS_TAG
+    script: ./.travis/deploy.sh hadoop_shred $TRAVIS_TAG
     skip_cleanup: true
     on:
       tags: true
-      condition: '"$(.travis/is_hadoop_shred_release_tag.sh $TRAVIS_TAG)" == "" && $? == 0'
+      condition: '"$(.travis/is_release_tag.sh hadoop_shred $TRAVIS_TAG)" == "" && $? == 0'
   - provider: script
-    script: ./.travis/deploy_hadoop_enrich.sh $TRAVIS_TAG
```
BenFradet: @alexanderdean Thanks, so this should have been deploy_hadoop_event_recovery.sh that's what I thought.

3-enrich/spark-enrich/README.md
```diff
@@ -17,20 +17,20 @@ The process can be run on [Amazon Elastic MapReduce] [emr] using the [EmrEtlRunn
...
 
     $ git clone git://github.com/snowplow/snowplow.git
-    $ cd 3-enrich/scala-hadoop-enrich
+    $ cd 3-enrich/spark-enrich
     $ sbt assembly
 
 The 'fat jar' is now available as:
 
-    target/snowplow-hadoop-etl-x.x.x.jar
+    target/snowplow-spark-etl-x.x.x.jar
```
rgabo: `sbt assembly` produces `target/scala-2.11/snowplow-spark-enrich-x.x.x.jar` (rightly so). I think the name of the jar should be consistent with the project name (`spark-enrich`).

BenFradet: nicely spotted, thanks :+1:

rgabo:
@BenFradet I got to a point where I successfully ran `spark-enrich` using `spark-submit` locally and bumped into the first issue:

`EnrichJob` tries to cache `GeoLiteCity.dat` locally and it is hardcoded to download to HDFS which is not available for `local` runs. For `local` Spark master, `file:///` would work just fine.

I can work around the issue by configuring `MaxMind` from S3 so that no downloading happens, but I still wanted to raise that Spark execution model now allows us to run enrichment locally and that means that HDFS/DistributedCache might have to be avoided.

BenFradet:
@rgabo we had a discussion regarding this specific topic with @alexanderdean here: https://github.com/snowplow/snowplow/pull/3077#pullrequestreview-20015413 and it seems that there is no way around DistributedCache for now.

rgabo:
@BenFradet yep, seemed to remember the discussion and I've reread it now. As long as `FileUtils` just passes `s3` and similar paths through, it is easy to work it around. I'll see if I can think about something more idiomatic in Spark that works nicely with the GeoIP enrichment. Btw the file is larger (~100MB) if you actually purchase it :wink:

rgabo:
Well, `IpLookupsEnrichment` errors out with `java.io.FileNotFoundException: ./ip_geo (No such file or directory)`, bubbled up from `IpLookups` because symlinking to HDFS/S3 still requires an underlying HDFS for the `DistributedCache.createSymlink` and `DistributedCache.addCacheFile` calls.

Hack-ish workaround is to copy the GeoIP database to `./ip_geo` for local runs.

Ideally there will be a better approach to distributing the file, but it's definitely out-of-scope for this PR, just documenting my findings.

rgabo:
@BenFradet it seems that Kryo serialization is configured in `EnrichJob`, but `Array[com.snowplowanalytics.snowplow.enrich.common.outputs.EnrichedEvent]` is not registered.

```
java.lang.IllegalArgumentException: Class is not registered: com.snowplowanalytics.snowplow.enrich.common.outputs.EnrichedEvent[]
Note: To register this class use: kryo.register(com.snowplowanalytics.snowplow.enrich.common.outputs.EnrichedEvent[].class);
        at com.esotericsoftware.kryo.Kryo.getRegistration(Kryo.java:488)
        at com.esotericsoftware.kryo.util.DefaultClassResolver.writeClass(DefaultClassResolver.java:97)
        at com.esotericsoftware.kryo.Kryo.writeClass(Kryo.java:517)
        at com.esotericsoftware.kryo.Kryo.writeClassAndObject(Kryo.java:622)
        at org.apache.spark.serializer.KryoSerializerInstance.serialize(KryoSerializer.scala:312)
        at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:324)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
        at java.lang.Thread.run(Thread.java:745)
```

Adding the missing class to `EnrichJob.classesToRegister` fixes the issue. I have successfully ran `spark-enrich` producing both `good` and `bad` output in one pass on local mode. YAY!

alexanderdean:
Thanks for all the QA support @rgabo - it's appreciated!

BenFradet:
@rgabo I added a few classes yesterday among them `Array[EnrichedEvent]` cf https://github.com/snowplow/snowplow/pull/3077/files#diff-9e5742dff28d1d166dcb990bcdaa5a79R56

rgabo:
@BenFradet no problem. I totally forgot to pull before I checked what's going on. Unfortunately if you force push, notifications don't work as you'd expect them as there are no "new" commits. I'd avoid force pushes to better document what's going on and to be outside collaborator friendly :wink:

rgabo:
Hi @BenFradet, I'm facing an interesting issue when running spark-enrich on EMR with a private MaxMind GeoIP database configured. It seems that the configured database is not properly download on the executors and even though it resides on S3, the Spark executors fail the enrichment with `Could not extract geo-location from IP address [x.x.x.x]: [java.io.FileNotFoundException: ./ip_geo (No such file or directory)`.

As we have a paid database, we would prefer to use that instead of the publicly available GeoLiteCity.dat baked into the fat jar.

Any pointers appreciated!

BenFradet:
@rgabo Thanks a lot for reporting it, I'll investigate why the symlink doesn't seem to be created.

rgabo:
Thanks @BenFradet, temporarily removing the `ip_lookups` enrichment altogether results in 100% good rows. Let me know if you find something.

rgabo:
@BenFradet I found an inefficiency during my testing: The `isEmpty` checks for `bad` and `good` RDDs result in a single task assigned to a single executor that will check all partitions for the existence of at least 1 item using `take(1) == 0`. The complexity is `O(n)` where n is the number of partitions, but the problem is that with `gz` files this means that a single executor has to download all filles, uncompress all files, process all files only to realize there are no bad rows. That is obviously inefficient :smile:

A simple paralellization of the `isEmpty` check does the trick: https://github.com/sspinc/snowplow/commit/d5509966f03be5603443d2fdae7637c254c1ee04

There's a lot more unrelated work on that branch that we need to do to maintain backward-compatibility, but feel free to cherry pick the commit itself.

snowplowcla:
@rgabo has signed the Software Grant and Corporate Contributor License Agreement

BenFradet:
@rgabo I was actually thinking about removing the empty checks entirely and write empty files similar to what is done currently in hadoop-{enrich,shred}

rgabo:
@BenFradet that was my thought as well, you'll have to adjust the tests as they are looking for the existence or absence of those directories.

BenFradet:
@rgabo I pushed a fix for the enrichments, I'll test it today and keep you updated.

rgabo:
@BenFradet thanks for the heads-up, good stuff.

BenFradet:
doesn't seem to have fixed the issue, I'll try a few other things.

BenFradet:
From what I've gathered so far, the `.dat` file is copied in HDFS but the symlink might not be created.

BenFradet:
@rgabo (finally) found a fix that is working on our side, added bonus: it doesn't rely on `DistributedCache`/HDFS anymore.

It'd be cool if you could also try it on your side :+1:

rgabo:
@BenFradet GREAT news, I will test later today or tomorrow.

rgabo:
@BenFradt I'm running the latest `spark-enrich` and see our GeoIPCity.dat cached:

```
17/03/17 14:49:11 INFO Utils: Fetching s3://secret-bucket/lib/snowplow/GeoIPCity.dat to /mnt/tmp/spark-4633d8c1-414f-405b-b036-00faadb62ece/userFiles-6a5bfc2c-609c-43eb-bd4b-f53a25e29f41/fetchFileTemp7566080315525470476.tmp
```

Unfortunately soon after the tasks fail with:

```
17/03/17 14:49:20 WARN TaskSetManager: Lost task 4.0 in stage 0.0 (TID 4, ip-10-64-33-234.ec2.internal, executor 1): java.nio.file.FileAlreadyExistsException: ./ip_geo
        at sun.nio.fs.UnixException.translateToIOException(UnixException.java:88)
        at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:102)
        at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:107)
        at sun.nio.fs.UnixFileSystemProvider.createSymbolicLink(UnixFileSystemProvider.java:457)
        at java.nio.file.Files.createSymbolicLink(Files.java:1043)
        at com.snowplowanalytics.snowplow.enrich.spark.EnrichJob$$anonfun$4$$anonfun$apply$3.apply(EnrichJob.scala:172)
        at com.snowplowanalytics.snowplow.enrich.spark.EnrichJob$$anonfun$4$$anonfun$apply$3.apply(EnrichJob.scala:169)
```

I am still troubleshooting if it somehow I am causing the issues, but I wanted to report the issue with the hope that you have some ideas.

Thanks!

BenFradet:
@rgabo I added https://github.com/snowplow/snowplow/pull/3077/files#diff-9e5742dff28d1d166dcb990bcdaa5a79R171 because of this but it's possible that another file is created between the deletion and creation.

I think I'm better off catching this exception and ignoring it.

BenFradet:
@rgabo pushed a fix, which I'll try in the afternoon

rgabo:
@BenFradet I've tested the change, all Spark executors fail a single task once but work on second retry and complete all tasks successfully. Definitely an improvement, now why would the first task for each executor fail...

rgabo:
Turns out it's not all executors, but in both test runs I've seen 8 tasks fail:

screen shot 2017-03-17 at 17 55 03

BenFradet:
Weird I haven't encountered those in my tests, I'd look into the corresponding logs to see why they failed.

rgabo:
@BenFradet what are next steps for Spark Enrich? I see that the features are scheduled for R89, is there anything missing? Is there any way there could be a "semi-official" release of the features that we could use and thoroughly test in anticipation of R89? Thanks!

/cc @alexanderdean

alexanderdean:
Hey @rgabo! We have R88 in final test; after the release of this, we will be rebasing R89 onto R88. At that point barring any issues that come up in the R89 QA, these jars will indeed be "semi-official"...

rgabo:
@alexanderdean awesome, looking forward to both.

BenFradet:
closing in favor of #3203

PR #3082, opened 2017-01-28
Update streaming components to latest vn (1.11.86)

acgray:
Also limiting deps to only necessary components to reduce build size


acgray:
This passes but still uses my private fork of Scalazon with the upgraded SDK version. I've submitted a PR with those changes and I'll update this PR once they get released...

snowplowcla:
@acgray has signed the Software Grant and Corporate Contributor License Agreement.

BenFradet:
@acgray 

Unfortunately, we''re moving away from Scalazon since it doesn't seem to be maintained anymore (cf #3300 and #3341). Moreover, the sdk versions will be updated as part of #3274.

Thanks a lot for your contribution anyway!

acgray:
No worries, that suits me just fine!

PR #3133, opened 2017-02-27
R88 Angkor Wat

chuwy:


5-data-modeling/event-manifest-populator/src/main/scala/com/snowplowanalytics/snowplow/eventpopulator/Utils.scala
```diff
@@ -0,0 +1,216 @@
...
+    */
+  case class DeduplicationTriple(eventId: String, fingerprint: String, etlTstamp: String)
+
+  /**
+    * Error happening when TSV doesn't match expected number of lines
+    *
+    * @param columns encountered number of columns
+    * @param line original event TSV line
+    */
+  case class ParsingError(columns: Int, line: String) extends RuntimeException {
```
alexanderdean: Not used anywhere

5-data-modeling/event-manifest-populator/src/main/scala/com/snowplowanalytics/snowplow/eventpopulator/Utils.scala
```diff
@@ -0,0 +1,216 @@
...
+    filtered.map(runId => runIdFormat.print(runId))
+  }
+
+  /**
+    * Get list of all folders inside S3 bucket with enriched archive
+    *
+    * @param jobConf parameters parsed from command line
+    * @return list of **dates** (without "run") of for arhives
+    */
+  private def ls3s(jobConf: JobConf): List[DateTime] = {
```
alexanderdean: What does ls3s stand for? I can't parse it.

5-data-modeling/event-manifest-populator/src/main/scala/com/snowplowanalytics/snowplow/eventpopulator/Utils.scala
```diff
@@ -0,0 +1,216 @@
...
+    def message: String = s"Wrong EnrichedEvent TSV columns number. Expected 130, got [$columns]. Line: [$line]"
+  }
+
+  /**
+    * Get run ids from `jobConf.enrichedInBucket` happened since `jobConf.since`
+    *
+    * @param jobConf input configuration
+    * @return list of timestamps (run ids)
+    */
+  def getRuns(jobConf: JobConf): List[String] = {
```
alexanderdean: Have you tested this @chuwy ? It's not working for me.

5-data-modeling/event-manifest-populator/src/main/scala/com/snowplowanalytics/snowplow/eventpopulator/Main.scala
```diff
@@ -0,0 +1,132 @@
...
+      case NonFatal(e) =>
+        val error = s"Error parsing --since option. It should conform YYYY-MM-dd-HH-mm-ss OR YYYY-MM-dd pattern.\n${e.toString}"
+        toProcMsgNel(error).failure
+    }
+
+  /**
+    * Drop s3 protocol and make sure there's trailing slash
+    */
+  private def normalizeBucket(bucket: String): String = {
+    val schemaless = bucket.drop(5) // drop s3:// prefix
```
alexanderdean: I strongly dislike this magic value of 5 - this should be the length of a constant string, "s3://", which should also used everywhere else that "s3://" is hard-coded...

5-data-modeling/event-manifest-populator/run.py
```diff
@@ -0,0 +1,281 @@
...
+from boto.emr.step import JarStep
+
+
+DIR_WITH_JAR = "./target/scala-2.11/"
+JAR_FILE = "event-manifest-populator-0.1.0-rc1.jar"
+
+S3_REGIONS = {'us-east-1': Location.DEFAULT,
+              'us-west-1': Location.USWest,
+              'us-west-2': Location.USWest2,
+              'eu-west-1': Location.EU,
```
alexanderdean: Replace Location.EU with "eu-west-1"

5-data-modeling/event-manifest-populator/run.py
```diff
@@ -0,0 +1,281 @@
...
+
+def run_emr(args):
+    validate_input_path(args.enriched_archive)
+    if args.since is not None:
+        validate_since(args.since)
+        since_arg = ["--since", args.since]
+    else:
+        since_arg = []
+
+    c = boto.connect_s3(profile_name=args.profile)
```
alexanderdean: Is args.profile optional, or not? Passing in profile_name=None to Boto is never a good idea.

5-data-modeling/event-manifest-populator/run.py
```diff
@@ -0,0 +1,281 @@
...
+    except ValueError:
+        try:
+            datetime.strptime(since, SINCE_PRECISE_FORMAT)
+        except ValueError:
+            print("Incorrect --since format. Should conform YY-MM-dd or YY-MM-dd-HH-mm-ss format")
+            sys.exit(1)
+
+
+def get_valid_region(location):
+    return S3_LOCATIONS[location]
```
alexanderdean: In my local install, this dictionary lookup fails with a KeyError because the `location` is a unicode string, but the map contains non-unicode keys...

5-data-modeling/event-manifest-populator/run.py
chuwy: It's optional. `None` is what it [gets by default](http://boto.cloudhackers.com/en/latest/ref/s3.html#boto.s3.connection.S3Connection) replacing it with default profile later. Do we need to make it required?

5-data-modeling/event-manifest-populator/src/main/scala/com/snowplowanalytics/snowplow/eventpopulator/Utils.scala
chuwy: `ls` stands for unix `ls` command.

5-data-modeling/event-manifest-populator/src/main/scala/com/snowplowanalytics/snowplow/eventpopulator/Main.scala
chuwy: Ok, we can replace it with `bucket.stripPrefix("s3://")`

5-data-modeling/event-manifest-populator/src/main/scala/com/snowplowanalytics/snowplow/eventpopulator/Utils.scala
chuwy: Yes. I just tested it again and see it works. There's also unit tests for filter.

chuwy: @alexanderdean I think I found a bug! How many runs were in the bucket you tested it in? S3 truncates keys to 1000 in list requests. If there were >1000 then it probably didn't reach specified date at all.

5-data-modeling/event-manifest-populator/run.py
chuwy: There's no difference between `u'location'` and `'location'` in python. What was location value?

5-data-modeling/event-manifest-populator/run.py
alexanderdean: Interesting. Nothing was working for me on Ubuntu. I'll rerun later this morning when there's a new version of this script and see what happens...

5-data-modeling/event-manifest-populator/run.py
alexanderdean: u'eu-west-1'. Thinking about this more - `Location.USWest2` is just a constant value for us-west-2 etc. So I think the get_valid_region function could be safely rewritten as:

```
if location = 'us-east-1' or unicode("us-east-1") then Location.DEFAULT else location
```

5-data-modeling/event-manifest-populator/src/main/scala/com/snowplowanalytics/snowplow/eventpopulator/Main.scala
alexanderdean: That's a start! Then define `val BucketPrefix = "s3://"` and use that everywhere in this file instead of the magic string.

5-data-modeling/event-manifest-populator/src/main/scala/com/snowplowanalytics/snowplow/eventpopulator/Utils.scala
alexanderdean: So `ls 3s`? I'm still not seeing it.

5-data-modeling/event-manifest-populator/src/main/scala/com/snowplowanalytics/snowplow/eventpopulator/Utils.scala
alexanderdean: Ah missing pagination, yes I buy that. Please fix this and I'll retry.

5-data-modeling/event-manifest-populator/run.py
chuwy: In Python 2:
```python
>>> 'us-east-1' == unicode('us-east-1')
True
```

In Python 3: `NameError: name 'unicode' is not defined`

It should be fine regardless unicode.

5-data-modeling/event-manifest-populator/src/main/scala/com/snowplowanalytics/snowplow/eventpopulator/Utils.scala
chuwy: Ah! It's `3s`!

5-data-modeling/event-manifest-populator/run.py
alexanderdean: This is the error I was getting:

```
  File "./run.py", line 153, in run_emr
    r = get_valid_region(jar_bucket.get_location())
  File "./run.py", line 110, in get_valid_region
    return S3_LOCATIONS[location]
KeyError: u'eu-west-1'
```

alexanderdean: To be clear, there are three issues here:

1. Issues with using unicode as a key lookup
2. Incorrect value for Location.EU
3. Missing regions from the dictionary

That's why my suggestion is to replace with some variant of:

```
if location = 'us-east-1' or unicode("us-east-1") then Location.DEFAULT else location
```

If you have an alternate fix for these three issues, let me know!

PR #3153, opened 2017-03-11
Gcp snowplow

colobas:
- [x] Add Pub/Sub as a sink for the Scala Stream Collector
- [x] Change the example hocon for the Scala Stream Collector, to include GCP configs
- [x] Add Pub/Sub as a sink for Stream Enrich
- [x] Add Pub/Sub as a source for Stream Enrich
- [x] Change the example hocon for Stream Enrich, to include GCP configs
- [x] Enable retrieval of enrichments and iglu resolvers from Cloud Datastore
- [ ] Add more detailed configurations for the sinks and sources (for example allowing Ack Policies other than the default)

### DIFF
--- a/2-collectors/scala-stream-collector/examples/config.hocon.sample
+++ b/2-collectors/scala-stream-collector/examples/config.hocon.sample
@@ -53,6 +53,7 @@ collector {
     # Sinks currently supported are:
     # 'kinesis' for writing Thrift-serialized records to a Kinesis stream
     # 'kafka' for writing Thrift-serialized records to kafka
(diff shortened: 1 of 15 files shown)


alexanderdean:
Thanks for all this @colobas! @chuwy - Gui has completed his internship now. Based on this, my thinking is that you do a review of this PR (there is no urgency to do this), and then we assign another engineer to apply the changes (so we basically get two sets of eyes on it). What do you think?

chuwy:
@alexanderdean I like it! I'll do a review by the end of this week.

BenFradet:
incorporated into #3624, thanks :+1:

PR #3235, opened 2017-05-19
Adding a replacement macro for the snowplow uuid within a redirect url.

rbolkey:
In order to perform ID syncing, we needed the ability to inject the snowplow id within a redirect url.

I decided on a replacement key of ${SP_UUID}. Not sure what kind of documentation needs to change.


rbolkey:
No need to merge this.

PR #3243, opened 2017-06-07
Feature/ga

BenFradet:
This is based on r89 so that we can get 2.11 and a recent sbt (plus it'll make for an easier rebase :smile:).

- [x] unstruct_event
- [ ] context
- [ ] events.x mapping
- [ ] composite fields
- [ ] let-through fields


3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/MeasurementProtocolAdapter.scala
```diff
@@ -0,0 +1,165 @@
...
+  val hitTypeData = Map(
+    "pageview" -> HitTypeData(
+      SchemaKey(vendor, "page_view", format, schemaVersion).toSchemaUri,
+      Map(
+        "dl" -> "documentLocationURL",
+        "dh" -> "documentHostName",
+        "dp" -> "documentPath",
+        "dt" -> "documentTitle"
+      )
+    ),
```
BenFradet: I haven't found a better way to maintain this mapping yet if someone has any ideas they're welcome.

BenFradet:
closing for now

PR #3273, opened 2017-07-04
Scala Stream Collector: allow raw query params

### LINKED ISSUE #3272: Scala Stream Collector: certain reserved characters result in a 400 response
At the moment certain reserved characters (such as `=`) when passed in a URL param result in a 400 currently from the collector. Other reserved characters such as `&`, `$` and `@` do not.

Ideally these characters should be urlencoded however in some limited environments where there may be restricted variable substitution (e.g., email platforms) it may be required to pass this data through. Snowplow should accept this data in raw form.

Example:
where user_id is the product of ```base64("test")```
```curl -XGET http://localhos:8080/i?e=pv&uid=dGVzdA==```
should return a 200 rather than a 400.

In addition the current setting for `uri-parsing-mode` in [config.hocon](https://raw.githubusercontent.com/snowplow/snowplow/master/2-collectors/scala-stream-collector/examples/config.hocon.sample) sits under spray.can.server whereas it should sit under spray.can.server.parsing according to [reference.conf](https://github.com/spray/spray/blob/master/spray-can/src/main/resources/reference.conf). I'm not certain on the behaviour of Spray but if it isn't currently picking up the setting in `config.hocon` due to the path then it may use the default value of strict.

miike:
Fixes #3272

### DIFF
--- a/2-collectors/scala-stream-collector/examples/config.hocon.sample
+++ b/2-collectors/scala-stream-collector/examples/config.hocon.sample
@@ -131,11 +131,11 @@ spray.can.server {
   # adds the 'Remote-Address' header to every request automatically.
   remote-address-header = on
 
-  uri-parsing-mode = relaxed
   raw-request-uri-header = on
 
   # Define the maximum request length (the default is 2048)
   parsing {
     max-uri-length = 32768
+    uri-parsing-mode = relaxed-with-raw-query
   }
 }


snowplowcla:
@miike has signed the Software Grant and Corporate Contributor License Agreement

miike:
@BenFradet My mistake, I was testing this locally and forgot to move the file. Fixed.

PR #3288, opened 2017-07-09
Data modeling: solve issue with incorrect timestamps breaking the web mode (closes #3276)

### LINKED ISSUE #3276: Data modeling: solve issue with incorrect timestamps breaking the web model
A rare edge case of the `derived_tstamp` has caused the web model to break. In the specific example, the `derived_tstamp` landed in Redshift as `26/1/1 15:0:59.9930`. 

To calculate the `time_engaged_in_s` the model applies an `EPOCH` function to the `derived_tstamp`. An `EPOCH` timestamp counts the seconds that have surpassed since 1/1/1970 therefore an error occurs when the timestamp populates with date 26/1/1.

We need to ensure all timestamps are `EPOCH` translatable for the current `time_engaged_in_s` function to work.

keanerobinson:


BenFradet:
The web model has moved to https://github.com/snowplow/web-data-model, please open this PR there.

PR #3314, opened 2017-07-19
Added ability to use custom partition keys for sinks

ashwinGokhale:


snowplowcla:
Thanks for your pull request. Is this your first contribution to a Snowplow open source project? Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).

📝 Please visit https://github.com/snowplow/snowplow/wiki/CLA to learn more and sign.

Once you've signed, please reply here (e.g. I signed it!) and we'll verify. Thanks.

ashwinGokhale:
I signed it!

snowplowcla:
Confirmed! @ashwinGokhale has signed the Software Grant and Corporate Contributor License Agreement. Thanks so much.

3-enrich/stream-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/kinesis/sources/AbstractSource.scala
```diff
@@ -191,7 +191,16 @@ abstract class AbstractSource(config: KinesisEnrichConfig, igluResolver: Resolve
       validatedMaybeEvent match {
         case Success(co) => (tabSeparateEnrichedEvent(co), if (config.useIpAddressAsPartitionKey) {
             co.user_ipaddress
-          } else {
+          } 
+	  else if (config.partitionKey != null) {
+          try
+            String.valueOf(co.getClass.getDeclaredField(config.partitionKey).get(co))
+          catch {
+            case _: Exception =>
```
chuwy: We never want to catch all exceptions. This should be something *similar* to 

```scala
  case _ @ (_: NoSuchFieldException | _: IllegalAccessException) => UUID.randomUUID.toString
```

But it is also very important to distinguish between cases where user tries to use non-existing property (must fail config validation long before runtime reaches this code) and when he tries to use `null` property.

3-enrich/stream-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/kinesis/sources/AbstractSource.scala
```diff
@@ -191,7 +191,16 @@ abstract class AbstractSource(config: KinesisEnrichConfig, igluResolver: Resolve
       validatedMaybeEvent match {
         case Success(co) => (tabSeparateEnrichedEvent(co), if (config.useIpAddressAsPartitionKey) {
             co.user_ipaddress
-          } else {
+          } 
+	  else if (config.partitionKey != null) {
+          try
+            String.valueOf(co.getClass.getDeclaredField(config.partitionKey).get(co))
```
chuwy: I'm not entirely sure it is going to work. As all JVM fields in Scala are private. Access to them is made only through getters.

3-enrich/stream-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/kinesis/sources/AbstractSource.scala
```diff
@@ -191,7 +191,16 @@ abstract class AbstractSource(config: KinesisEnrichConfig, igluResolver: Resolve
       validatedMaybeEvent match {
         case Success(co) => (tabSeparateEnrichedEvent(co), if (config.useIpAddressAsPartitionKey) {
```
chuwy: This whole `if-else` expression can be extracted as separate function with tests.

3-enrich/stream-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/kinesis/KinesisEnrichApp.scala
```diff
@@ -378,4 +378,5 @@ class KinesisEnrichConfig(config: Config) {
   val maxBackoff = backoffPolicy.getLong("maxBackoff")
 
   val useIpAddressAsPartitionKey = outStreams.hasPath("useIpAddressAsPartitionKey") && outStreams.getBoolean("useIpAddressAsPartitionKey")
+  val partitionKey = if (outStreams.hasPath("partitionKey")) outStreams.getString("partitionKey") else null
```
chuwy: Why not `Option[String]`?

3-enrich/stream-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/kinesis/sources/AbstractSource.scala
chuwy: Would also like to see some tests if this is going to be extracted into separate function.

BenFradet:
Closing since this is a dupe of #3362 

@ashwinGokhale please try to keep only one PR open for the same issue.

PR #3350, opened 2017-07-28
NSQ Support for Scala Stream Collector and Stream Enrich

aldemirenes:
I've bumped scala version to 2.11 in the stream enrich because NSQ client library requires it. I think bumping the version will be done in another ticket. I can rebase the project or create new PR after bumping the version.


BenFradet:
@aldemirenes please don't open a new pull request every time, just push your changes to the branch against which the pr is open

aldemirenes:
@BenFradet Sure Ben, sorry for missing about this.

3-enrich/stream-enrich/project/BuildSettings.scala
```diff
@@ -22,9 +22,9 @@ object BuildSettings {
     organization          :=  "com.snowplowanalytics",
     version               :=  "0.10.0",
     description           :=  "The Snowplow Enrichment process, implemented as an Amazon Kinesis app",
-    scalaVersion          :=  "2.10.1",
+    scalaVersion          :=  "2.11.8",
```
BenFradet: this will be taken care of separately

3-enrich/stream-enrich/project/BuildSettings.scala
```diff
@@ -22,9 +22,9 @@ object BuildSettings {
...
     scalacOptions         :=  Seq("-deprecation", "-encoding", "utf8",
-                                  "-feature", "-target:jvm-1.7"),
+                                  "-feature", "-target:jvm-1.8"),
```
BenFradet: same

3-enrich/stream-enrich/project/Dependencies.scala
```diff
@@ -36,20 +36,21 @@ object Dependencies {
...
-    val igluClient           = "0.4.0"
+    val scalaz7              = "7.0.9"
+    val igluClient           = "0.5.0"
     val snowplowTracker      = "0.3.0"
     // Scala (test only)
     // val specs2               = "2.3.6" Conflicts with com.chuusai:shapeless
-    val specs2               = "2.2.3"
-    val scalazSpecs2         = "0.1.2"
+    val specs2               = "2.3.11"
+    val scalazSpecs2         = "0.2"
```
BenFradet: please don't change the dependencies other than the ones you need, this will be taken care of separately

3-enrich/stream-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/kinesis/KinesisEnrichApp.scala
```diff
@@ -344,6 +347,15 @@ class KinesisEnrichConfig(config: Config) {
...
   val kafkaBrokers = kafka.getString("brokers")
 
+  private val nsq = enrich.getConfig("NSQ")
+  val nsqGoodSourceTopicName = nsq.getString("good-source-topic")
+  val nsqGoodSourceChannelName = nsq.getString("good-source-channel")
+  val nsqGoodSinkTopicName = nsq.getString("good-sink")
+  val nsqBadTopicName = nsq.getString("bad-sink")
+  val nsqHost = nsq.getString("nsq-host")
+  val nsqdPort = nsq.getInt("nsqd-port")
+  val nsqlookupdPort = nsq.getInt("nsqlookupd-port")
```
BenFradet: let's rename this property to `nsq-lookupd-port`

3-enrich/stream-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/kinesis/KinesisEnrichApp.scala
```diff
@@ -344,6 +347,15 @@ class KinesisEnrichConfig(config: Config) {
   private val kafka = enrich.getConfig("kafka")
   val kafkaBrokers = kafka.getString("brokers")
 
+  private val nsq = enrich.getConfig("NSQ")
+  val nsqGoodSourceTopicName = nsq.getString("good-source-topic")
+  val nsqGoodSourceChannelName = nsq.getString("good-source-channel")
+  val nsqGoodSinkTopicName = nsq.getString("good-sink")
+  val nsqBadTopicName = nsq.getString("bad-sink")
+  val nsqHost = nsq.getString("nsq-host")
+  val nsqdPort = nsq.getInt("nsqd-port")
```
BenFradet: let's rename to `nsq-port`

3-enrich/stream-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/kinesis/sinks/NsqSink.scala
```diff
@@ -0,0 +1,54 @@
...
+// Java
+import java.nio.charset.StandardCharsets.UTF_8
+
+// NSQ
+import com.github.brainlag.nsq.NSQProducer
+
+
+/**
+  * Stdouterr Sink for Scala enrichment
+  */
```
BenFradet: please modify the scaladoc according to your code

3-enrich/stream-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/kinesis/sinks/NsqSink.scala
```diff
@@ -0,0 +1,54 @@
...
+    case InputType.Good => config.nsqGoodSinkTopicName
+    case InputType.Bad => config.nsqBadTopicName
+  }
+
+  /**
+    *
+    * @param events Sequence of enriched events and (unused) partition keys
+    * @return Whether to checkpoint
+    */
+
```
BenFradet: empty line

3-enrich/stream-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/kinesis/sinks/NsqSink.scala
```diff
@@ -0,0 +1,54 @@
...
+  override def storeEnrichedEvents(events: List[(String, String)]): Boolean = {
+    events foreach {
+      e => producer.produce(topicName, e._1.getBytes(UTF_8))
```
BenFradet: why not use [`produceMulti`](https://github.com/brainlag/JavaNSQClient/blob/master/src/main/java/com/github/brainlag/nsq/NSQProducer.java#L72) ?

3-enrich/stream-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/kinesis/sources/NsqSource.scala
```diff
@@ -0,0 +1,70 @@
...
+        val bytes = Base64.decodeBase64(msg.getMessage())
+        enrichAndStoreEvents(List(bytes))
+        msg.finished()
+      }
+    }
+
+    // use NSQLookupd
+    val lookup = new DefaultNSQLookup
+    lookup.addLookupAddress(config.nsqHost, config.nsqlookupdPort)
+    val consumer = new NSQConsumer(lookup, config.nsqGoodSourceTopicName, config.nsqGoodSourceChannelName, nsqCallback)
```
BenFradet: why not use the consumer using an error callback: https://github.com/brainlag/JavaNSQClient/blob/master/src/main/java/com/github/brainlag/nsq/NSQConsumer.java#L57?

2-collectors/scala-stream-collector/examples/config.hocon.sample
```diff
@@ -92,6 +93,23 @@ collector {
...
+      good-sink: "{{collectorNsqSourceGoodTopicName}}"
+
+      #Topic name for bad event sink
+      bad-sink: "{{collectorNsqSinkTopicBadName}}"
+        
+      #Host name for NSQ tools
+      nsq-host: "{{NsqHost}}"
+
+      #Port for nsqd
+      nsqd-port: "{{NsqdPort}}"
```
aldemirenes: Actually there are two different tools for NSQ, one of them is nsqd and other is nsqlookupd. These two tools should run on the same machine therefore I used one host name which is nsq-host for both of them. Since they are two different tool, both of them have different port number.

BenFradet: Agreed but it is just confusing as is, let's have nsq-port and nsq-lookup-port

3-enrich/stream-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/kinesis/sinks/NsqSink.scala
aldemirenes: Every event string must be converted to the byte array. After converting event, I send them with "produce". I did not see any reason to get together them again and sent with produceMulti. Is there any reason which you think using produceMulti is better solution ?

BenFradet: it's only one message sent to nsq instead of n: https://github.com/brainlag/JavaNSQClient/blob/master/src/main/java/com/github/brainlag/nsq/NSQCommand.java#L97-L109

3-enrich/stream-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/kinesis/KinesisEnrichApp.scala
aldemirenes: Since tool name is nsqlookupd, I think it is better in this way. What you think ?

BenFradet: Now that I think about it, it's pointless to repeat nsq, since the globbing config is already called, let's refactor it as:

```config
NSQ {
  host = ...
  port = ...
  lookup-port = ...
}
```

3-enrich/stream-enrich/examples/config.hocon.sample
```diff
@@ -41,6 +43,27 @@ enrich {
     secret-key: "iam"
   }
 
+  # config for NSQ sink and NSQ source
+NSQ {
+  #Topic name for good event source
```
BenFradet: nit: could you add a space after # here and below?

3-enrich/stream-enrich/examples/config.hocon.sample
```diff
@@ -41,6 +43,27 @@ enrich {
...
+     
+  #Host name for NSQ tools
+  host: "{{NsqHost}}"
+
+  #Port for nsqd
+  port: "{{NsqdPort}}"
+
+  #Port for nsqlookupd
+  lookup-port: {{NsqlookupdPort}}
+}
```
BenFradet: could you indent this section properly?

3-enrich/stream-enrich/examples/config.hocon.sample
```diff
@@ -41,6 +43,27 @@ enrich {
...
+  # config for NSQ sink and NSQ source
+NSQ {
+  #Topic name for good event source
+  good-source-topic: "{{sinkKinesisNsqSourceGoodTopicName}}"
+
+  #Channel name for good event source
+  good-source-channel: "{{sinkKinesisNsqSourceGoodChannelName}}"
+
+  #Topic name for bad event sink
+  bad-sink: "{{sinkKinesisNsqSinkTopicBadName}}"
```
BenFradet: for the 3 properties above, why aren't you using the configuration in `streams`, i.e.:

- `in.raw`
- `out.enriched`
- `out.bad`

?

3-enrich/stream-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/kinesis/sources/NsqSource.scala
```diff
@@ -0,0 +1,96 @@
...
+  /**
+    * Consumer will be started to wait new message.
+    */
+  override def run(): Unit = {
+
+    val nsqCallback = new NSQMessageCallback {
+      override def message(msg: NSQMessage) = {
+        val bytes = msg.getMessage()
+        enrichAndStoreEvents(List(bytes))
+        msg.finished()
```
BenFradet: `enrichAndStoreEvents` gives back a boolean saying whether or not you should checkpoint which, in the case of NSQ, means acking the message.

2-collectors/scala-stream-collector/examples/config.hocon.sample
```diff
@@ -103,6 +119,7 @@ collector {
     }
 
     # Incoming events are stored in a buffer before being sent to Kinesis/Kafka.
+    # Note: Buffering is not supported by NSQ; will be ignored
```
BenFradet: Can't we use `byte-limit` as [output buffer size](https://github.com/brainlag/JavaNSQClient/blob/master/src/main/java/com/github/brainlag/nsq/NSQConfig.java#L76-L78) and `time-limit` as [output buffer timeout](https://github.com/brainlag/JavaNSQClient/blob/master/src/main/java/com/github/brainlag/nsq/NSQConfig.java#L84-L86)?

3-enrich/stream-enrich/examples/config.hocon.sample
aldemirenes: There are channel fields in the NSQ additionally. I thought that keeping channel and topic name separated can be confusing therefore I put all of the topic and channel fields under NSQ section.

BenFradet: but this introduces two sources of truth and people won't know what is used so I suggest reusing everything and keeping the channel config here

2-collectors/scala-stream-collector/examples/config.hocon.sample
aldemirenes: I think buffer related fields are put in the NSQConfig only. I can not find anywhere which they are used.

BenFradet: what do you mean?

- http://nsq.io/clients/tcp_protocol_spec.html#identify
- https://github.com/brainlag/JavaNSQClient/blob/master/src/main/java/com/github/brainlag/nsq/NSQCommand.java#L44-L54
- https://github.com/brainlag/JavaNSQClient/blob/9f5d767eb5e3495ebc032eae711fba7929fed132/src/main/java/com/github/brainlag/nsq/Connection.java#L72

aldemirenes: Sorry, since I did not see the direct usage of these field, I thought they are not used. I will make necessary changes then.

aldemirenes: > **output_buffer_size (nsqd v0.2.21+)** the size in bytes of the buffer nsqd will use when writing to this client.
This part is from [here](http://nsq.io/clients/tcp_protocol_spec.html)
As I understand, these buffer fields are used while nsqd is writing to the clients. I think there is no buffering option for producer. It is obvious from [here](https://github.com/brainlag/JavaNSQClient/blob/master/src/main/java/com/github/brainlag/nsq/NSQProducer.java#L28) too because NSQProducer class does not take any NSQConfig argument and creates default NSQConfig which buffer fields are null in the default NSQConfig.  I will add buffering to stream-enrich but I think this is not possible for stream-collector. What do you think ?

BenFradet: I don't think it really applies to anything since the buffer settings are producer-related and not consumer-related so they shouldn't be in stream-enrich either.

PR #3362, opened 2017-08-07
Feature/issue-1924

ashwinGokhale:


snowplowcla:
@ashwinGokhale has signed the Individual Contributor License Agreement. Thanks so much

alexanderdean:
> most people won't go through the trouble of looking at the EnrichedEvent class to find out which property they want to use

No, we would point them at the official docs for the enriched event model:

https://github.com/snowplow/snowplow/wiki/canonical-event-model

The limiting factor here I think is more that people might want to partition on e.g. a user identity in a custom context, and this PR wouldn't let you do this. Still, I think this is a good first phase of functionality.

3-enrich/stream-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/kinesis/sources/AbstractSource.scala
```diff
@@ -173,13 +174,23 @@ abstract class AbstractSource(config: KinesisEnrichConfig, igluResolver: Resolve
...
   }
 
+  def getProprertyValue(co: EnrichedEvent, property: String): String = {
+    val info = Introspector.getBeanInfo(co.getClass)
+    info.getPropertyDescriptors.foreach { p =>
+      if (p.getName.equals(property))
+        return String.valueOf(p.getReadMethod.invoke(co))
+    }
+    UUID.randomUUID().toString
+  }
```
BenFradet: could we refactor this as:

```scala
def getProprertyValue(ee: EnrichedEvent, property: String): String =
  Introspector
    .getBeanInfo(ee.getClass)
    .getPropertyDescriptors
    .find(_.getName == property)
    .map(p => String.valueOf(p.getReadMethod.invoke(ee)))
    .getOrElse(UUID.randomUUID().toString
```

3-enrich/stream-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/kinesis/sources/AbstractSource.scala
```diff
@@ -173,13 +174,23 @@ abstract class AbstractSource(config: KinesisEnrichConfig, igluResolver: Resolve
...
+
   /**
    * Convert incoming binary Thrift records to lists of enriched events
    *
    * @param binaryData Thrift raw event
    * @return List containing successful or failed events, each with a
    *         partition key
    */
+
```
BenFradet: empty line

3-enrich/stream-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/kinesis/KinesisEnrichApp.scala
```diff
@@ -378,4 +378,12 @@ class KinesisEnrichConfig(config: Config) {
   val maxBackoff = backoffPolicy.getLong("maxBackoff")
 
   val useIpAddressAsPartitionKey = outStreams.hasPath("useIpAddressAsPartitionKey") && outStreams.getBoolean("useIpAddressAsPartitionKey")
+  val partitionKey:Option[String] = if (outStreams.hasPath("partitionKey") && outStreams.getString("partitionKey").isEmpty) {
+    Some(null)
```
BenFradet: I'm keen on treating an empty partition key as `None`

ashwinGokhale: A user might want a null partition key for keyless messaging, so if the partitionKey property is specified, but the value is an empty string, the key will be null.

BenFradet: I don't think this has any benefit over a random UUID since specifying no partition will result in a round robin and hashing UUIDs will result in hopefully the same thing.

BenFradet:
I still don't like this approach of using reflection and finding _some_ property based on a name, the main reason being it's really expensive: we're reflecting and we're going over all fields of `EnrichedEvent` for every single event we enrich.

Surely, most of the fields wouldn't make good candidates for a partition key and a limited number would. My current thinking is still around suggesting a limited set of potential fields that would make good partition keys.

alexanderdean:
Let's come up with a shortlist

BenFradet:
Here's my shortlist, tell me what you think @alexanderdean :+1: 

- `event_id`
- `event_fingerprint`
- `domain_userid`
- `network_userid`
- `user_ipaddress` we should remove useIpAddressAsPartitionKey as part of this PR btw
- `domain_sessionid`
- `user_fingerprint`

One thing worth investigating would be the proportion of events where those fields are missing

alexanderdean:
List looks good @BenFradet !

BenFradet:
@ashwinGokhale Could we refactor this PR to leverage this list and not reflection?

ashwinGokhale:
@BenFradet Sure

3-enrich/stream-enrich/examples/config.hocon.sample
```diff
@@ -70,6 +70,10 @@ enrich {
       enriched: "{{enrichStreamsOutEnriched}}"
       bad: "{{enrichStreamsOutBad}}"
 
+      # See https://github.com/snowplow/snowplow/wiki/canonical-event-model for possible partition keys
+      # If "partitionKey" property is empty, the partition key will be null
```
BenFradet: could you we modify the documentation accordingly?

3-enrich/stream-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/kinesis/KinesisEnrichApp.scala
```diff
@@ -378,4 +378,12 @@ class KinesisEnrichConfig(config: Config) {
   val maxBackoff = backoffPolicy.getLong("maxBackoff")
 
   val useIpAddressAsPartitionKey = outStreams.hasPath("useIpAddressAsPartitionKey") && outStreams.getBoolean("useIpAddressAsPartitionKey")
```
BenFradet: could we remove this option as a result of your changes?

BenFradet:
Integrated into #3274 , thanks!

PR #3108, opened 2017-02-09
Updated atomic.events column encodings to optimize compression

clarkeandrew:
These column encodings are based on the recommendations from running ANALYZE COMPRESSION on our atomic.events table (1042240319 rows). The resulting table within redshift is ~50% of the size of the original table.

PR #3370, opened 2017-08-14
Update atomic.events to use ZSTD compression

miike:
This updates the canonical `atomic.events` model to use ZSTD compression by default on all columns except for the `collector_tstamp` SORTKEY as recommended by AWS. This global compression seems to perform best across the board (though individual implementations may differ on what ANALYZE COMPRESSION outputs due to data variance).

I've bumped this to 0.9.0 as although the data is completely backwards compatible it is not possible to alter compression on an existing table (however new installations can immediately use this compression). This is likely to need some companion documentation for users upgrading from previous versions of the table to 0.9.0.

Ref #3108


alexanderdean:
Many thanks for this @miike ! We are looking into ZSTD very seriously now, so this is great timing.

snowplowcla:

alexanderdean:
When merging into the release branch, make sure to update the commit message...

PR #3372, opened 2017-08-14
Scala Common Enrich: bump mandrill event versions to 1-0-1

acgray:
See https://github.com/snowplow/iglu-central/pull/628


snowplowcla:
@acgray has signed the Software Grant and Corporate Contributor License Agreement

BenFradet:
Could we add this and snowplow/iglu-central#628 to Knossos ? @alexanderdean @knservis

alexanderdean:
No, Knossos has been kicked off already. Per the off-site, we are trying not to add tickets to sprints that have already started.

BenFradet:
Yup, that's my bad I wasn't aggressive enough in scheduling, scheduling now.

BenFradet:
mmmh I thought those would in 102...

BenFradet:
Hey @acgray ,

Apparently we have customers running the Mandrill integration without running into issues with this new subaccount property.

Because updating the jsonpaths associated with the 1-0-1 schemas would fail their data loading into redshift if we don't update their tables, we're postponing this feature.

The commit now lives in #3716, closing.

alexanderdean:
I don't think this is blocked - we can emit the newer schemas, they will pass validation, we simply will be loading the older versions (1-0-0s) into Redshift.

BenFradet:
true...reopening

Issue #3483, opened 2017-10-29
Scala Common Enrich: default port to 443 if scheme is https


miike:
The current behaviour of `explodeUri` sets port to 80 if there is no explicit port number in the URL. To correct this if no port number is present and the scheme is https the port returned should be 443.

[Related discourse post](https://discourse.snowplowanalytics.com/t/bug-js-tracker-is-logging-port-numbers-incorrectly/1573)

alexanderdean:
"Default" -> "default"

PR #3484, opened 2017-10-29
If scheme is https and no explicit port default port to 443

miike:
Returns port 443 if no explicit port is set and scheme = 'https'

Related issue #3483


snowplowcla:

PR #3499, opened 2017-11-08
Issue 2792

knservis:
Fixed the test


3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/OlarkAdapter.scala
```diff
@@ -121,7 +122,7 @@ object OlarkAdapter extends Adapter {
               }
             }
           } catch { 
-            case e: Exception => {
+            case NonFatal(e) => {
               val exception = JU.stripInstanceEtc(e.toString).orNull
```
BenFradet: same `getMessage` here

PR #3510, opened 2017-11-16
Issue 3508

knservis:


3-enrich/spark-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.spark/bad/CorruptedThriftLinesSpec.scala
```diff
@@ -27,16 +27,19 @@ class CorruptedThriftLinesSpec extends Specification with EnrichJobSpec {
   override def appName = "corrupted-thrift-lines"
   sequential
   "A job which processes a corrupted input line" should {
-    runEnrichJob(getClass().getResource("CorruptedThriftLinesSpec.line.lzo").toString(), "thrift",
-      "1", false, List("geo"), false, false, false, false)
+    if (!isLzoSupported) "native-lzo not supported" in skipped
```
BenFradet: let's use https://github.com/snowplow/snowplow/blob/master/2-collectors/scala-stream-collector/src/test/scala/com.snowplowanalytics.snowplow.collectors.scalastream/utils/SplitBatchSpec.scala#L36

knservis: Is there a way to include a reason the test was skipped with that syntax?

BenFradet: iirc the runner tells you why it didn't run, needs to be verified though

PR #3545, opened 2017-12-15
R98 Argentomagus

BenFradet:
- dev
  - [x] rebase once r97 is merged
  - [x] nsq client 1.2.0
  - [x] nsq client 1.2.0-M1 -> 1.2.0
- qa
  - [x] cbor
  - [x] crossdomain
  - [x] redirect macro
  - [x] bouncing scheme
  - [x] raw location header
  - [x] 22017 collector date
  - [x] kinesis server side encryption
  - [x] nsq
  - [x] load testing
- admin tasks
  - [x] wiki
  - [x] blogpost
  - [x] merge wiki
  - [x] dates (wiki, bp, cl)

### DIFF
--- a/2-collectors/scala-stream-collector/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
+++ b/2-collectors/scala-stream-collector/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
@@ -121,7 +121,8 @@ class CollectorService(
       )
 
     val (httpResponse, badRedirectResponses) = buildHttpResponse(
-      event, partitionKey, queryParams, headers.toList, redirect, pixelExpected, bounce, config.streams.sink)
+      event, partitionKey, queryParams, headers.toList, redirect, pixelExpected, bounce,
+      config.streams.sink, config.redirectMacro)
     (httpResponse, badRedirectResponses ++ sinkResponses)
   }
 
@@ -206,10 +207,11 @@ class CollectorService(
     redirect: Boolean,
     pixelExpected: Boolean,
     bounce: Boolean,
-    sinkConfig: SinkConfig
+    sinkConfig: SinkConfig,
+    redirectMacroConfig: RedirectMacroConfig
   ): (HttpResponse, List[Array[Byte]]) =
     if (redirect) {
-      val (r, l) = buildRedirectHttpResponse(event, partitionKey, queryParams)
+      val (r, l) = buildRedirectHttpResponse(event, partitionKey, queryParams, redirectMacroConfig)
       (r.withHeaders(r.headers ++ headers), l)
     } else {
       sinkConfig match {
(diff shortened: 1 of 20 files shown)


alexanderdean:
I love the name!

2-collectors/scala-stream-collector/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
```diff
@@ -133,13 +139,22 @@ class CollectorService(
...
+      .map { c =>
+        HttpResponse(
+          entity = HttpEntity(
+            contentType = ContentType(MediaTypes.`text/xml`, HttpCharsets.`ISO-8859-1`),
+            string = s"""<?xml version=\"1.0\"?>\n<cross-domain-policy>
+                        |  <allow-access-from domain=\"${c.domain}\" secure=\"${c.secure}\" />
+                        |</cross-domain-policy>""".stripMargin
+          )
+        )
+      }.getOrElse(HttpResponse(404, entity = "404 not found"))
```
chuwy: I think pattern-matching would be more readable here.

2-collectors/scala-stream-collector/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
```diff
@@ -133,13 +139,22 @@ class CollectorService(
         `Access-Control-Allow-Headers`("Content-Type")
       ))
 
+  override def flashCrossDomainPolicy: HttpResponse =
```
chuwy: I thought we don't use `override` unless we're overriding behavior? In `Service` this method is empty.

BenFradet: I tend to always write `override`, this way you know this method was first defined elsewhere (here in the trait).

Here it's not the case but if this was still abstract you have the added benefit of signature checking between this and the overriden method instead of defining a new method.

2-collectors/scala-stream-collector/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
BenFradet: good point 👍

BenFradet:
I don't know wtf is going on with travis but publishing errors out for stream enrich without error messages...

https://travis-ci.org/snowplow/snowplow/jobs/322933438#L3324

BenFradet:
mmh same thing happened for spark enrich: https://travis-ci.org/snowplow/snowplow/jobs/315961847#L3998

chuwy:
Similar thing happened for Iglu recently - build was interrupting silently because of `exit(1)` in script and error message wasn't in logs. Travis disappoints me these days.

alexanderdean:
Looks like we are waiting on @knservis' code review now

PR #3546, opened 2017-12-15
Scala Stream Collector: replace Location header with RawHeader to preserve double encoding

rbolkey:
Making this change in order preserver double encoding URLs for use in
redirect chains. This also matches the behavior of the Clojure Collector.


snowplowcla:
@rbolkey has signed the Software Grant and Corporate Contributor License Agreement. Thanks so much

PR #3585, opened 2018-01-18
GDPR-1 release Milestone 149

knservis:

PR #3588, opened 2018-01-19
Scala Common Enrich: add PII Enrichment

knservis:
Sorry I could not maintain the comments (as the previous PR was closed before its time).

Please keep formatting comments for the autoformatter PR.  

This is now ready @chuwy @BenFradet. Can you guys please have a look?

### DIFF
--- a/3-enrich/stream-enrich/build.sbt
+++ b/3-enrich/stream-enrich/build.sbt
@@ -14,7 +14,7 @@ lazy val root = project.in(file("."))
   .settings(
     organization  :=  "com.snowplowanalytics",
     name          :=  "snowplow-stream-enrich",
-    version       :=  "0.13.0",
+    version       :=  "0.14.0",
     description   :=  "The Snowplow Enrichment process, implemented as an Amazon Kinesis app",
     scalaVersion  :=  "2.11.11",
     scalacOptions :=  BuildSettings.compilerOptions,
--- a/3-enrich/stream-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/SpecHelpers.scala
+++ b/3-enrich/stream-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/SpecHelpers.scala
@@ -39,7 +39,7 @@ object SpecHelpers {
   /**
    * The Stream Enrich being used
    */
-  val EnrichVersion = "kinesis-0.13.0-common-0.31.0"
+  val EnrichVersion = "kinesis-0.14.0-common-0.31.0"
 
   val TimestampRegex = "[0-9\\s-:.]+"
 


3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
```diff
@@ -0,0 +1,342 @@
...
+
+  /**
+   * Strategy for this field
+   *
+   * @return PiiStrategy a strategy to be applied to this field
+   */
+  def strategy: PiiStrategy
+
+  /**
+   * The POJO mutator for this field
```
BenFradet: This is for pojo and json now, no?

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
```diff
@@ -0,0 +1,342 @@
...
+            case JObject(context) => modifyObjectIfSchemaMatches(context)
+            case x                => x
+          }))
+        case JField("data", JObject(unstructEvent)) =>
+          ("data", modifyObjectIfSchemaMatches(unstructEvent))
+      }))
+    else null
+
+  private def modifyObjectIfSchemaMatches(context: List[(String, json4s.JValue)]): JObject = {
+    val fields: Map[String, JValue] = List("schema", "data").flatMap(k => context.toMap.get(k).map(v => (k -> v))).toMap
```
BenFradet: I have trouble understanding why you have to go through all this, isn't the above equivalent to:

```scala
context.toMap.filterKeys(List("schema", "data").contains)
```

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
```diff
@@ -0,0 +1,342 @@
...
+
+  private def getHashFunction(strategyFunction: String): Validation[String, MessageDigest] =
+    try {
+      MessageDigest.getInstance(strategyFunction).success
+    } catch {
+      case e: NoSuchAlgorithmException =>
+        s"Could not parse PII enrichment config: ${e.getMessage()}".failure
+    }
+
+  private def extractFields(piiFields: List[JObject], strategy: PiiStrategy) =
```
BenFradet: return type

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
```diff
@@ -0,0 +1,342 @@
...
+  private def modifyObjectIfSchemaMatches(context: List[(String, json4s.JValue)]): JObject = {
+    val fields: Map[String, JValue] = List("schema", "data").flatMap(k => context.toMap.get(k).map(v => (k -> v))).toMap
+    fields
+      .get("schema")
+      .flatMap(s =>
+        SchemaKey.parse(s.extract[String]).map(schemaCriterion.matches).toOption match {
+          case Some(matches) if matches =>
+            fields.get("data").flatMap(d => Some(JObject(List(("schema", s), ("data", jsonPathReplace(d))))))
+          case _ => None
+      })
```
BenFradet: curious if scalafmt is going to catch this, this is just a marker for me for later :D

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
knservis: json is just another field (it contains a json string instead of a scalar) in the POJO, so at this level they are both POJO mutators the way I see it. Maybe the we should rename the Pojo mutator to "scalar mutator" to make that clear, What do you think?

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
knservis: good point

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
knservis: Is this about "=>" alignment? If so it will not align them unless they are in consecutive lines and it needs to break the line here as it would be longer that the maximum configured.

BenFradet: no it's about indentation {} not being on the same level as ()

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
BenFradet: it would be even better to extract out the List or do s => s == "schema" || s == "data", that would avoid having to create n lists.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
BenFradet: I kinda prefer scalar mutator but I don't mind either way.

knservis: scalar mutator sounds more accurate to me. I am changing it.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
knservis: 👍

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
knservis: I tried splitting       }) via a newline to see if it will indent them differently and it joined them back. There must be a setting for that.

knservis:
@BenFradet All good for you here?

BenFradet:
yes, I'll let @chuwy do another pass 👍

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
```diff
@@ -0,0 +1,342 @@
...
+    }
+
+  private def extractFields(piiFields: List[JObject], strategy: PiiStrategy): Validation[String, List[PiiField]] =
+    piiFields.map {
+      case JObject(List(("pojo", JObject(List(("field", JString(fieldName))))))) =>
+        ScalarMutators
+          .get(fieldName)
+          .map(PiiScalar(strategy, _).success)
+          .getOrElse(s"The specified pojo field ${fieldName} is not supported".failure)
+      case JObject(List(("json", jsonField))) =>
```
chuwy: Feel free to ignore, but I'd like to not make a precedent of pattern-matching JSONs this way. It's quite dangerous if for example junior developer who doesn't know that `keys` in `JObject(keys)` is order-sensetive.

knservis: @chuwy It's easy for anyone to miss that especially if you haven't used json4s. In this case there is a single value and this is very convenient, but if there is a better, clearer way to get that value out I would like to know.

alexanderdean:
In a separate track, can we:

1. Rebase
2. Treat this as definitely R100
3. Come up with an archeological site

knservis:
@alexanderdean 

1. Done for this and the target branch. 
2. Done on target branch (https://github.com/snowplow/snowplow/pull/3585/commits/feafbc7e29073072279a0198adad254954ab0346 on https://github.com/snowplow/snowplow/pull/3585)
3. Epidaurus (https://en.wikipedia.org/wiki/Epidaurus)

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
```diff
@@ -0,0 +1,350 @@
...
+ * @param schemaCriterion the schema for which the strategy will be applied
+ * @param jsonPath the path where the strategy will be applied
+ */
+final case class PiiJson(strategy: PiiStrategy, fieldMutator: PiiConstants.Mutator, schemaCriterion: SchemaCriterion, jsonPath: String)
+    extends PiiField {
+  implicit val json4sFormats = DefaultFormats
+
+  override def applyStrategy(fieldValue: String): String =
+    if (fieldValue != null)
+      compact(render(parse(fieldValue).transformField {
```
chuwy: I'm worry that this function will be applied to all fields in JSON.

```scala
parseJson("""{"foo": 1, "nested": {"foo": 2}}""").transformField {
  case JField("foo", JInt(x)) => JField("bar", JInt(x + 1))
}
JObject(List(("bar", JInt(2)), ("nested", JObject(List(("bar", JInt(3)))))))
```

So, having a context with `schema` and `data` keys seems a bit dangerous for me now. As far as I can understand it won't transform anything if structure is not exactly as configured. Still - maybe this is wasteful to apply this recursive function to all fields in JSON?

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
```diff
@@ -0,0 +1,350 @@
...
+    else null
+
+  private def modifyObjectIfSchemaMatches(context: List[(String, json4s.JValue)]): JObject = {
+    val fields: Map[String, JValue] = context.toMap.filterKeys(s => s == "schema" || s == "data")
+    fields
+      .get("schema")
+      .flatMap(s =>
+        SchemaKey.parse(s.extract[String]).map(schemaCriterion.matches).toOption match {
+          case Some(matches) if matches =>
+            fields.get("data").flatMap(d => Some(JObject(List(("schema", s), ("data", jsonPathReplace(d))))))
```
chuwy: `Some` == `pure`;
`flatMap ∘ pure` == `map`;

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
```diff
@@ -0,0 +1,350 @@
...
+          ("data", modifyObjectIfSchemaMatches(unstructEvent))
+      }))
+    else null
+
+  private def modifyObjectIfSchemaMatches(context: List[(String, json4s.JValue)]): JObject = {
+    val fields: Map[String, JValue] = context.toMap.filterKeys(s => s == "schema" || s == "data")
+    fields
+      .get("schema")
+      .flatMap(s =>
+        SchemaKey.parse(s.extract[String]).map(schemaCriterion.matches).toOption match {
```
chuwy: Having recursive nature of `transformField` it turns out that if we'll get a context like:

```
{
  "schema": "iglu:com.acme/event/jsonschema/1-0-0",
  "data": {"schema": 1, "data": 2, "foo": 3}
}
```

then `s.extract[String]` will fail at runtime? If that is the case (I'd like to have a test for that assumption) then above request about of changing `transformWith` is even stronger.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
knservis: What you are saying is true, and another reason for me to not like json4s, however there is a guard in modifyObjectIfSchemaMatches, which is it only applies the modification if the schema matches. The only documented way that I could find on modifying a field does exactly that (in http://json4s.org/) and yes it does modify fields in multiple levels using pattern matching.  Using json4s, I am not sure what is a good alternative to avoid looking at all the fields recursively.

knservis: NM ^^^

chuwy:
```scala
def transformObject(json: JValue, f: PartialFunction[JField, JField]): JValue = json match {
  case JObject(fields) =>
    JObject(fields.map { field => if (f.isDefinedAt(field)) f(field) else field })
  case other => other
}
```

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
```diff
@@ -0,0 +1,350 @@
...
+          }))
+        case JField("data", JObject(unstructEvent)) =>
+          ("data", modifyObjectIfSchemaMatches(unstructEvent))
+      }))
+    else null
+
+  private def modifyObjectIfSchemaMatches(context: List[(String, json4s.JValue)]): JObject = {
+    val fields: Map[String, JValue] = context.toMap.filterKeys(s => s == "schema" || s == "data")
+    fields
+      .get("schema")
```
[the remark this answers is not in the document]
knservis: Even though I did change that bit it was not to use iglu core. Can you give me an idea on how it would be used here?

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
```diff
@@ -0,0 +1,354 @@
...
+final case class PiiJson(strategy: PiiStrategy, fieldMutator: PiiConstants.Mutator, schemaCriterion: SchemaCriterion, jsonPath: String)
+    extends PiiField {
+  implicit val json4sFormats = DefaultFormats
+
+  override def applyStrategy(fieldValue: String): String =
+    if (fieldValue != null) {
+      compact(render(parse(fieldValue) match {
+        case JObject(jObject) => {
+          val jObjectMap = jObject.toMap
+          val updated = jObjectMap.filterKeys(s => s == "data").mapValues {
```
chuwy: `_ == "data"`

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
```diff
@@ -0,0 +1,354 @@
...
+    } else {
+      "Schema key %s is not supported. A '%s' enrichment must have schema '%s'."
+        .format(schemaKey, supportedSchema.name, supportedSchema)
+        .failure
+    }
+}
+
+/**
+ * The PiiPseudonymizerEnrichment runs after all other enrichments to find fields that are configured as PII (personally
+ * identifiable information) and apply some anonymization (currently only psudonymizantion) on them. Currently a single
```
chuwy: `s/psudonymizantion/pseudonymization`

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
```diff
@@ -0,0 +1,354 @@
...
+        else s"PII Configuration: pii field does not include 'pojo' nor 'json' fields. Got: [${compact(field)}]".failure
+      }
+      case json => s"PII Configuration: pii field does not contain an object. Got: [${compact(json)}]".failure
+    }.sequenceU
+
+  private def extractPiiScalarField(strategy: PiiStrategy, fieldName: String): Validation[String, PiiScalar] =
+    ScalarMutators
+      .get(fieldName)
+      .map(PiiScalar(strategy, _).success)
+      .getOrElse(s"The specified pojo field ${fieldName} is not supported".failure)
```
chuwy: Unnecessary curly braces.

3-enrich/scala-common-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/utils/TestResourcesRepositoryRef.scala
```diff
@@ -0,0 +1,66 @@
...
+import com.fasterxml.jackson.core.JsonParseException
+import com.fasterxml.jackson.databind.JsonNode
+import com.github.fge.jackson.JsonLoader
+
+// Scalaz
+import scalaz.Scalaz._
+
+//Snowplow
+import com.snowplowanalytics.iglu.client.repositories.{RepositoryRef, RepositoryRefConfig}
+import com.snowplowanalytics.iglu.client.{SchemaKey,                  Validated, utils, validation}
```
chuwy: Weird spacing.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
```diff
@@ -0,0 +1,354 @@
...
+      }))
+    } else null
+
+  private def modifyObjectIfSchemaMatches(context: List[(String, json4s.JValue)]): JObject = {
+    val fieldsObj = context.toMap
+    (for {
+      schema <- fieldsObj.get("schema")
+      parsedSchemaMatches <- SchemaKey.parse(schema.extract[String]).map(schemaCriterion.matches).toOption
+      data <- fieldsObj.get("data")
+      if(parsedSchemaMatches)
```
chuwy: `if parsedSchemaMatches`

knservis:
Thanks @chuwy I took care of the last comments too.

PR #3590, opened 2018-01-22
Add HQL file for creating atomic.events table

miike:
This adds a Hive (Athena flavoured) definition for atomic.events. There's already a `hive-storage` directory so I'm not sure if it belongs in there but this is distinctly AWS Athena flavoured.

It makes the following assumptions:
- Data is stored in S3
- Data is partitioned by a timestamp labelled "partition"
- Data is stored in Parquet format and compressed in Snappy format

These properties can be changed by modifying the last 4 lines of this file.


miike:
That's a fair question and I'm not too sure. I'm open to any suggestions on how I could relax those assumptions to make it more inclusive?

chuwy:
I think Hive/Athena Loader has a right to exist in a form similar to our Snowflake Loader.

alexanderdean:
Thanks for the PR @miike!

> those are strong assumptions and that makes me wonder if it should live in the repository or elsewhere, e.g. a guide of some sort

> Hive/Athena Loader has a right to exist

I agree - for me the missing "piece in the middle" is the Snowplow BAPHM Loader (Blob storage, Avro, Parquet, Hive Metastore, sorry for the 🧀 ). This is discussed further in this Discourse thread:

https://discourse.snowplowanalytics.com/t/snowflake-db-other-storage-targets/1721/5?u=alex

The other thing this PR has raised is that we really need to crack on with the enriched event refactor. The later we leave it, the more loaders and storage targets there are to migrate...

snowplowcla:

BenFradet:
I think this should live as a guide outside of the repo, thanks @miike , feel free to share it on discourse

alexanderdean:
Yep, HQL isn't something we want to add formal support for now, especially given the upcoming atomic.events refactor.

PR #3618, opened 2018-02-14
PII identification events

knservis:


3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
```diff
@@ -248,97 +339,134 @@ object PiiPseudonymizerEnrichment extends ParseableEnrichment {
...
-      parsedSchemaMatches <- SchemaKey.parse(schema.extract[String]).map(schemaCriterion.matches).toOption
+      schema <- fieldsObj.get("schema")
+      schemaStr = schema.extract[String]
+      parsedSchemaMatches <- SchemaKey.parse(schemaStr).map(schemaCriterion.matches).toOption
       data                <- fieldsObj.get("data")
       if parsedSchemaMatches
-    } yield JObject(fieldsObj.updated("schema", schema).updated("data", jsonPathReplace(data)).toList)).getOrElse(JObject(context))
+      updated = jsonPathReplace(data, strategy, schemaStr)
+    } yield
+      (JObject(fieldsObj.updated("schema", schema).updated("data", updated._1).toList), updated._2)).getOrElse((JObject(context), List()))
```
BenFradet: can we break this down, e.g.:

```scala
(modifiedJson, modifiedFields) = jsonPathReplace()
// ...
yield {
  val updatedFields = fieldsObj
    .updated("schema", schema)
    .updated("data", modifiedJson)
    .toList
  (JObject(updatedFields), modifiedFields))
    .getOrElse((JObject(context), List.empty[JsonModifiedField]))
}
```

you might find better variable names than above

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
```diff
@@ -248,97 +339,134 @@ object PiiPseudonymizerEnrichment extends ParseableEnrichment {
...
   }
 
   // Configuration for JsonPath
   private val JsonPathConf =
-    Configuration.builder().options(JOption.SUPPRESS_EXCEPTIONS).jsonProvider(new JacksonJsonNodeJsonProvider()).build()
+    Configuration
+      .builder()
+      .options(JOption.SUPPRESS_EXCEPTIONS, JOption.ALWAYS_RETURN_LIST)
```
BenFradet: can you explain your choice in a comment

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
```diff
@@ -248,97 +339,134 @@ object PiiPseudonymizerEnrichment extends ParseableEnrichment {
...
+      .jsonProvider(new JacksonJsonNodeJsonProvider())
+      .build()
 
   /**
    * Replaces a value in the given context data with the result of applying the strategy that value.
```
BenFradet: this comment might need refreshing given the function seems to have changed significantly

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
```diff
@@ -351,3 +479,69 @@ case class PiiStrategyPseudonymize(hashFunction: MessageDigest) extends PiiStrat
...
+/**
+ * Case class for capturing JSON field modifications.
+ */
+case class JsonModifiedField(field: String, originalValue: String, modifiedValue: String, jsonPath: String, schema: String)
+    extends ModifedField
+
+/**
+ * Parent class for classes that serialize the values that were modified during the PII enrichment.
+ */
+case class PiiModifiedFields(val modifiedFields: PiiConstants.ModifiedFields, val strategy: PiiStrategy)
```
BenFradet: can we make all those final? I don't feel good about publicly-exposed non final case class

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
```diff
@@ -351,3 +479,69 @@ case class PiiStrategyPseudonymize(hashFunction: MessageDigest) extends PiiStrat
...
+
+/**
+ * Custom serializer for PiiStrategy class
+ */
+class PiiStrategySerializer
```
BenFradet: can we restrict those serializers' scope?

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
```diff
@@ -351,3 +479,69 @@ case class PiiStrategyPseudonymize(hashFunction: MessageDigest) extends PiiStrat
...
+    extends CustomSerializer[PiiStrategy](format =>
+      ({
+        case jo: JObject =>
+          implicit val json4sFormats = DefaultFormats
```
BenFradet: why do we need another formats inside the serializer?

BenFradet: same thing below

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
```diff
@@ -351,3 +479,69 @@ case class PiiStrategyPseudonymize(hashFunction: MessageDigest) extends PiiStrat
...
+          ("schema" -> PiiTransformationSchema) ~
+            ("data" ->
+              ("pii" -> decompose(
+                pmf.modifiedFields
+                  .map {
+                    case s: ScalarModifiedField => "pojo" -> s
+                    case j: JsonModifiedField   => "json" -> j
+                  }
+                  .groupBy(_._1)
+                  .mapValues(_.map(_._2))
```
BenFradet: can't we extract this data wrangling from the json construction? this will make thing easier to understand

BenFradet: also can we turn this into a fold, that'll be more performant

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
```diff
@@ -334,25 +451,39 @@ final case class PiiJson(strategy: PiiStrategy,
...
+    val modifiedFields  = MutableList[JsonModifiedField]()
+    documentContext.map(jsonPath,
+                        ScrambleMapFunction(strategy, modifiedFields, fieldMutator.fieldName, jsonPath, schema))
+    (JsonMethods.fromJsonNode(documentContext.json[JsonNode]), modifiedFields.toList)
   }
 }
 
-case class ScrambleMapFunction(val strategy: PiiStrategy) extends MapFunction {
+case class ScrambleMapFunction(strategy: PiiStrategy,
+                               modifiedFields: MutableList[JsonModifiedField],
```
chuwy: I feel extremely bad about putting mutable data structures into `case class`es, especially non-private ones. I think this should `private class` at least.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
```diff
@@ -51,80 +54,166 @@ import common.utils.ScalazJson4sUtils
 import common.outputs.EnrichedEvent
 
 object PiiConstants {
-  type Mutator = (EnrichedEvent, String => String) => Unit
+  type ModifiedFields  = List[ModifedField]
```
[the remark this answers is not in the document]
knservis: Not sure what you mean by moving the declarations up in this case. Can you be more specific?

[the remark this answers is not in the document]
BenFradet: I meant moving https://github.com/snowplow/snowplow/pull/3618/files#diff-326e31f4255bc69c2c5c60cd68ec4104R510 up in the file.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
```diff
@@ -50,80 +53,166 @@ import common.utils.ScalazJson4sUtils
 import common.outputs.EnrichedEvent
 
 object PiiConstants {
```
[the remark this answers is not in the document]
knservis: Not a bad idea about the pii. The File was getting crowded and wanted a good solution to move all these constructs out. Pii seems like a good idea

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
knservis: As far as I can tell from bench marking this is not more performant. I tried your example and it was actually slower.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
```diff
@@ -367,3 +498,73 @@ case class PiiStrategyPseudonymize(hashFunction: MessageDigest) extends PiiStrat
...
+/**
+ * Parent class for classes that serialize the values that were modified during the PII enrichment.
+ */
+case class PiiModifiedFields(val modifiedFields: PiiConstants.ModifiedFields, val strategy: PiiStrategy)
+
+/**
+ * Custom serializer for PiiStrategy class
+ */
+class PiiStrategySerializer
+    extends CustomSerializer[PiiStrategy](format =>
```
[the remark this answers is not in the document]
knservis: It is not but it should be used

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
```diff
@@ -367,3 +498,73 @@ case class PiiStrategyPseudonymize(hashFunction: MessageDigest) extends PiiStrat
...
+      }, {
+        case psp: PiiStrategyPseudonymize =>
+          "pseudonymize" -> ("hashFunction" -> psp.hashFunction.getAlgorithm)
+      }))
+
+/**
+ * Custom serializer for PiiModifiedFields class
+ */
+class PiiModifiedFieldsSerializer
+    extends CustomSerializer[PiiModifiedFields](format => {
```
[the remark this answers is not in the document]
knservis: As above

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
```diff
@@ -223,7 +318,7 @@ object PiiPseudonymizerEnrichment extends ParseableEnrichment {
             .getOrElse(s"The specified json field ${compact(jsonField)} is not supported".failure)) |@|
```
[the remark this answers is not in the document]
knservis: Probably. If you get the whole field, you may have more information which could be useful in a long config, but having only the name would make a better error message. Changing it.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
BenFradet: Have you read:

>I reliably get those results https://gist.github.com/BenFradet/13c4a33ea6c0a850de8728f4cca3d82c
>anyway doing one pass instead of two over the data is preferable imo
>plus you avoid creating useless stuff

BenFradet: Anyways ... you can even save another pass by doing:

```scala
pmf.modifiedFields.foldLeft(Map.empty[String, List[ModifiedField]]) { case (m, mf) =>
  mf match {
    case s: ScalarModifiedField => m + ("pojo" -> (s :: m.getOrElse("pojo", List.empty[MF])))
    case j: JsonModifiedField => m + ("json" -> (j :: m.getOrElse("json", List.empty[MF])))
  }
}
```

knservis:
@BenFradet @chuwy I believe I fixed those issues. Have another look if you can. thanks

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/PiiPseudonymizerEnrichment.scala
knservis: This comment was lost here due to the update. I just found it in my emails

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentRegistry.scala
```diff
@@ -58,6 +57,7 @@ import registry.{
 }
 import registry.apirequest.{ApiRequestEnrichment, ApiRequestEnrichmentConfig}
 import registry.sqlquery.{SqlQueryEnrichment, SqlQueryEnrichmentConfig}
+import registry.pii.PiiPseudonymizerEnrichment
```
BenFradet: doesn't scalafmt handle reordering imports alphabetically?

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/pii/PiiPseudonymizerEnrichment.scala
```diff
@@ -0,0 +1,285 @@
...
+}
+
+/**
+ * The PiiPseudonymizerEnrichment runs after all other enrichments to find fields that are configured as PII (personally
+ * identifiable information) and apply some anonymization (currently only pseudonymization) on them. Currently a single
+ * strategy for all the fields is supported due to the configuration format, and there is only one implemented strategy,
+ * however the enrichment supports a strategy per field.
+ *
+ * The user may specify two types of fields POJO or JSON. A POJO field is effectively a scalar field in the
+ * EnrichedEvent, whereas a JSON is a "context" formatted field and it can be wither a scalar in the case of
```
BenFradet: can be a scalar

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/pii/Serializers.scala
```diff
@@ -0,0 +1,70 @@
...
+              ("pii" -> decompose(
+                pmf.modifiedFields.foldLeft(Map.empty[String, List[ModifiedField]]) {
+                  case (m, mf) =>
+                    mf match {
+                      case s: ScalarModifiedField =>
+                        m + ("pojo" -> (s :: m.getOrElse("pojo", List.empty[ModifiedField])))
+                      case j: JsonModifiedField => m + ("json" -> (j :: m.getOrElse("json", List.empty[ModifiedField])))
+                    }
+                }
+              ))            ~
```
BenFradet: is this scalfmt's doing? looks weird to me

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/pii/package.scala
```diff
@@ -0,0 +1,111 @@
...
+  type ApplyStrategyFn = (String, PiiStrategy) => (String, ModifiedFields)
+  type MutatorFn       = (EnrichedEvent, PiiStrategy, ApplyStrategyFn) => ModifiedFields
+
+  val JsonMutators   = Mutators.JsonMutators
+  val ScalarMutators = Mutators.ScalarMutators
+
+  // Configuration for JsonPath
+  private[pii] val JacksonNodeJsonObjectMapper = {
+    val objectMapper = new ObjectMapper()
+    objectMapper.configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false)
```
BenFradet: can we comment why this configuration was chosen, same below with `SUPPRESS_EXCEPTIONS`

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentRegistry.scala
knservis: I don't know all the settings, but I think it only sorts lists of imports from the same package. (e.g. import blah.{a, b, c} )

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/pii/PiiPseudonymizerEnrichment.scala
knservis: you mean for the types. Actually in the config they are called pojo and json. I can add the clarification.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/pii/Serializers.scala
knservis: We have asked scalafmt to align all `~` tokens (see .scalafmt.conf). It seems token here is aligned with something.. I am not sure what. I will try to see what happens if I remove the `~` from scalafmt config

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/pii/PiiPseudonymizerEnrichment.scala
BenFradet: no I meant the "wither" typo

knservis: Oh I see, it was supposed to be `either`

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentRegistry.scala
BenFradet: ah ok then we might look into something that does it for us too

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/pii/Serializers.scala
knservis: The result is this:
```
-              ))            ~
+              )) ~
```

knservis:
@chuwy Did you have a chance to have a look?

chuwy:
@knservis, nope sorry, will have a look right now.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/pii/PiiPseudonymizerEnrichment.scala
```diff
@@ -0,0 +1,301 @@
...
+
+  implicit val json4sFormats = DefaultFormats
+
+  override val supportedSchema =
+    SchemaCriterion("com.snowplowanalytics.snowplow.enrichments", "pii_enrichment_config", "jsonschema", 2, 0, 0)
+
+  def parse(config: JValue, schemaKey: SchemaKey): ValidatedNelMessage[PiiPseudonymizerEnrichment] = {
+    for {
+      conf <- matchesSchema(config, schemaKey)
+      enabled = ScalazJson4sUtils.extract[Boolean](conf, "enabled").toOption.getOrElse(false)
```
chuwy: Can you please clarify how `enabled` works here?

As far as I can tell, `EnrichmentRegistry.buildEnrichmentConfig` checks this value before and if it is `false` then `PiiPseudonymizerEnrichment.parse` will never be invoked. Am I missing something? Also it looks strange to me that on `enabled = false` below, we still have an enrichment object, which is just no-op.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/pii/PiiPseudonymizerEnrichment.scala
```diff
@@ -0,0 +1,301 @@
...
+      .toValidationNel
+    val validatedNel = (mutator |@| schemaCriterion |@| jsonPath)(PiiJson.apply)
+    validatedNel.leftMap(x => s"Unable to extract PII JSON: ${x.list.mkString(",")}")
+  }
+
+  private def getJsonMutator(fieldName: String): Validation[String, Mutator] =
+    JsonMutators
+      .get(fieldName)
+      .map(_.success)
+      .getOrElse(s"The specified json field ${fieldName} is not supported".failure)
```
chuwy: Unnecessary curly braces.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/pii/PiiPseudonymizerEnrichment.scala
```diff
@@ -0,0 +1,301 @@
...
+    ScalazJson4sUtils
+      .extract[String](config, "parameters", "strategy", "pseudonymize", "hashFunction")
+      .leftMap(_.getMessage)
+
+  private def matchesSchema(config: JValue, schemaKey: SchemaKey): Validation[String, JValue] =
+    if (supportedSchema.matches(schemaKey)) {
+      config.success
+    } else {
+      "Schema key %s is not supported. A '%s' enrichment must have schema '%s'."
+        .format(schemaKey, supportedSchema.name, supportedSchema)
```
chuwy: Quite strange that we use Scala interpolation everywhere and Java-ish `.format` in this single place.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/pii/PiiPseudonymizerEnrichment.scala
```diff
@@ -0,0 +1,301 @@
...
+ * @param schemaCriterion the schema for which the strategy will be applied
+ * @param jsonPath the path where the strategy will be applied
+ */
+final case class PiiJson(fieldMutator: Mutator, schemaCriterion: SchemaCriterion, jsonPath: String) extends PiiField {
+  implicit val json4sFormats = DefaultFormats
+
+  override def applyStrategy(fieldValue: String, strategy: PiiStrategy): (String, ModifiedFields) = {
+    val modifiedFields = MutableList[JsonModifiedField]()
+    if (fieldValue != null) {
+      (compact(render(parse(fieldValue) match {
```
chuwy: I think @BenFradet already raised that it's very hard to read this lisp-ish notation. Can we split this into separate values (and preferably smaller functions).

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/pii/PiiPseudonymizerEnrichment.scala
```diff
@@ -0,0 +1,301 @@
...
+object PiiPseudonymizerEnrichment extends ParseableEnrichment {
+
+  implicit val json4sFormats = DefaultFormats
+
+  override val supportedSchema =
+    SchemaCriterion("com.snowplowanalytics.snowplow.enrichments", "pii_enrichment_config", "jsonschema", 2, 0, 0)
+
+  def parse(config: JValue, schemaKey: SchemaKey): ValidatedNelMessage[PiiPseudonymizerEnrichment] = {
+    for {
+      conf <- matchesSchema(config, schemaKey)
```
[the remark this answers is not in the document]
knservis: Which alignment are you referring to in the above code?

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/pii/PiiPseudonymizerEnrichment.scala
```diff
@@ -0,0 +1,301 @@
...
+                                   PiiStrategyPseudonymize(hashFunctionName, hashFunction))
+      else
+        PiiPseudonymizerEnrichment(List(),
+                                   emitIdentificationEvent = false,
+                                   PiiStrategyPseudonymize(hashFunctionName, hashFunction))
+  }.leftMap(_.toProcessingMessageNel)
+
+  private[pii] def getHashFunction(strategyFunction: String): Validation[String, DigestFunction] =
+    strategyFunction match {
+      case "MD2" => { (b: Array[Byte]) =>
```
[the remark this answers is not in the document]
knservis: @chuwy This doesn't seem to work:
```
iiPseudonymizerEnrichment.scala:88: ambiguous reference to overloaded definition,
[error] both method md2Hex in object DigestUtils of type (x$1: String)String
[error] and  method md2Hex in object DigestUtils of type (x$1: java.io.InputStream)String
[error] match expected type ?
[error]       case "MD2"     => { DigestUtils.md2Hex }.success
```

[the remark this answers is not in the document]
chuwy: Yep, seems `.success` is the problem. You can use `case "MD2" => { DigestUtils.md2Hex _ }.success` or leave as is.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/pii/PiiPseudonymizerEnrichment.scala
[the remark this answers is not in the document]
chuwy: First two lines of `for`-comprehension. They are not aligned, while everything after `=` aligned.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/pii/PiiPseudonymizerEnrichment.scala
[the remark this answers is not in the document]
knservis: I don't think that was the problem:
```
PiiPseudonymizerEnrichment.scala:82: ambiguous reference to overloaded definition,
[error] both method md2Hex in object DigestUtils of type (x$1: String)String
[error] and  method md2Hex in object DigestUtils of type (x$1: java.io.InputStream)String
[error] match expected type ?
[error]       case "MD2"     => { DigestUtils.md2Hex _ }.success
```
The errors seem to suggest that because `md2Hex` is polymorphic the type of the argument is required.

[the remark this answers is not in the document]
chuwy: Ah, ok. So final `case "MD2"     => { DigestUtils.md2Hex(_: String) }.success` won't be much better than original solution. Up to you.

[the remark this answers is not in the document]
knservis: `case "MD2" => { DigestUtils.md2Hex(_: Array[Byte]) }.success` is better. Thanks

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/pii/PiiPseudonymizerEnrichment.scala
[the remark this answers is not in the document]
knservis: the `<-` token is aligned in an unbroken run but because there is an intervening `=` there are effectively three aligned blocks. Formatters are not as clever as to fully understand the language syntax I don't think. For all those cases one can use:
```
// format off
...
// format on
```

knservis:
@chuwy @BenFradet thanks guys for the PR. Is anything else remaining @chuwy ?

chuwy:
@knservis nope, greenlight from!

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/pii/PiiPseudonymizerEnrichment.scala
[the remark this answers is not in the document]
chuwy: Yep. That is exactly an argument to not force automated formatters - we're swapping one visual garbage with another.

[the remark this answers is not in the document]
knservis: In my opinion you probably get better overall results using a formatter to ensure consistent readability when a project gets above a few KLOC than not, even if you occasionally have to manually manage the format. It does most of the work for you, ensures consistency, and minimises formatting work so that people can focus on structure and potential defects. Consistent formatting is not a self-cause the way I see it but a means to an end, and that end is to improve maintainability and better discern potential defects. If teams can minimise the time spent on tasks that can be automated, is that not a win?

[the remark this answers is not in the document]
chuwy: > In my opinion you probably get better overall results using a formatter

I never said anything against using a formatter. I said only about **forcing** formatting.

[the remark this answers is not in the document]
alexanderdean: I agree with @knservis - it's better to apply global formatting and then use `format off / on` flags around pieces which are beyond the formatter's capability. This is a non-argument in other language communities (see e.g. Golang).

I don't agree that `// format off and // format on` is visual garbage - if anything, it's helpful - it says "here be dragons": the syntactical complexity of this piece of code is beyond a computer's ability to visualize it well for a human. That's interesting information.

On automated formatting - think of it as cruise control on a car - just because you have to sometimes switch it off for a complicated piece of driving, nobody seriously thinks it's better not to have it at all.

knservis:
@alexanderdean Do we need any changes here for your green tick?

PR #3624, opened 2018-02-19
R101 Neapolis

BenFradet:


alexanderdean:
Should be "add" not "Add"

alexanderdean:
Hey @BenFradet - I think this is likely to be R102 now.

BenFradet:
can't deploy rcs :fireworks: , logged https://github.com/travis-ci/dpl/issues/778

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
```diff
@@ -255,7 +240,7 @@ class CollectorService(
     queryParams.get("u") match {
       case Some(target) =>
         val canReplace = redirectMacroConfig.enabled && event.isSetNetworkUserId
-        val token = redirectMacroConfig.placeholder.getOrElse("${SP_NUID}")
+        val token = redirectMacroConfig.placeholder.getOrElse(s"$${SP_NUID}")
```
knservis: 👍

BenFradet:
This is just for symmetry, we don't do `AWSKinesis` or `AmazonKinesis` so `GooglePubSub`, or even  `GoogleCloudPubSub`, seemed clumsy to me, but I don't have strong preferences.

knservis:
@BenFradet Agreed not a great name (maybe GPubSub? ) but to me (again this may just be me) it would make things more intuitively readable, that is I would not need to go and verify for everything that I read, that it really is a specific product that variables are referring to and not some super class of all pub sub systems or some such generalisation.

BenFradet:
couldn't they come up with some original name... :smile:

alexanderdean:
Yes - I agree with @knservis - we should be going with GooglePubSub or similar, because PubSub is a category not a brand name (unlike Kafka or Kinesis)...

3-enrich/stream-enrich/build.sbt
```diff
@@ -10,47 +10,93 @@
...
-      Dependencies.Libraries.pureconfig,
-      Dependencies.Libraries.snowplowRawEvent,
-      Dependencies.Libraries.snowplowCommonEnrich,
-      Dependencies.Libraries.igluClient,
-      Dependencies.Libraries.snowplowTracker,
-      // Test
-      Dependencies.Libraries.specs2,
-      Dependencies.Libraries.scalacheck
-    )
+    buildInfoKeys := Seq[BuildInfoKey](organization, name, version,
```
chuwy: Didn't know about this plugin. Do you think we should standardize on it for all code bases? (instead of custom `scalify` settings)

2-collectors/scala-stream-collector/pubsub/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/PubSubCollector.scala
```diff
@@ -0,0 +1,44 @@
...
+package collectors
+package scalastream
+
+import scalaz._
+import Scalaz._
+
+import model._
+import sinks.PubSubSink
+
+object PubSubCollector extends App with Collector {
```
chuwy: I'd advise against `App` trait in any application https://www.reddit.com/r/scala/comments/4zvelw/app_trait_vs_main_method/

3-enrich/stream-enrich/build.sbt
BenFradet: yes there are a few tickets already open: https://github.com/snowplow/snowplow/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+buildinfo

2-collectors/scala-stream-collector/pubsub/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/PubSubSink.scala
```diff
@@ -0,0 +1,143 @@
...
+      .setRetryDelayMultiplier(backoffPolicy.multiplier)
+      .setTotalTimeout(Duration.ofMillis(backoffPolicy.totalBackoff))
+      .setInitialRpcTimeout(Duration.ofSeconds(10))
+      .setRpcTimeoutMultiplier(2)
+      .setMaxRpcTimeout(Duration.ofSeconds(10))
+      .build()
+
+  /** Checks that a PubSub topic exists **/
+  private def topicExists(projectId: String, topicName: String): \/[Throwable, Boolean] = for {
+    topicAdminClient <- toEither(Try(TopicAdminClient.create()))
```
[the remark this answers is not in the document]
BenFradet: what I wanted was to write the minimal amount of code that we'd have to throw away once we migrate to 2.12 where we can just replace every `\/` by `Either` but I do agree...

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/util.scala
```diff
@@ -0,0 +1,30 @@
...
+package com.snowplowanalytics.snowplow.collectors.scalastream
+
+import scala.util.{Failure, Try, Success}
+
+import scalaz._
+import Scalaz._
+
+object util {
+  // to rm once 2.12 as well as the right projections
+  def fold[A, B](t: Try[A])(ft: Throwable => B, fa: A => B): B = t match {
```
knservis: `ft: Throwable => B` looks a bit weird to me. Is the purpose of the provided function to use a throwable and return just B?

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/util.scala
```diff
@@ -0,0 +1,30 @@
...
+ * governing permissions and limitations there under.
+ */
+package com.snowplowanalytics.snowplow.collectors.scalastream
+
+import scala.util.{Failure, Try, Success}
+
+import scalaz._
+import Scalaz._
+
+object util {
```
knservis: would this work for this use: 
```
import scala.util.{ Either, Failure, Left, Right, Success, Try }

implicit def eitherToTry[A <: Exception, B](either: Either[A, B]): Try[B] = {
  either match {
    case Right(obj) => Success(obj)
    case Left(err) => Failure(err)

  }
}
implicit def tryToEither[A](obj: Try[A]): Either[Throwable, A] = {
  obj match {
    case Success(something) => Right(something)
    case Failure(err) => Left(err)
  }
}
```
(source: https://stackoverflow.com/questions/22532357/either-to-try-and-vice-versa-in-scala)

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/util.scala
BenFradet: that's just how [fold](https://www.scala-lang.org/api/2.12.4/scala/util/Try.html#fold[U](fa:Throwable=%3EU,fb:T=%3EU):U) works

chuwy:
> can't deploy rcs 🎆 , logged travis-ci/dpl#778

Same for me, QA is blocked.

PR #3646, opened 2018-03-01
Pii emit, hocon, integration test and autoformatting

knservis:
This PR needs a bit of work still:

- [x] Verify that new curly braces in hocon do not affect anything ```sink = {{ sinkType }}``` and ```    source = {{ sourceType }}``` with @BenFradet 
- [x] Add scalafmt to stream-enrich to make more readable.


knservis:
@BenFradet can you please verify that the hocon changes don't cause any issue, and that scalafmt is something we want here?

BenFradet:
@knservis do you mean https://github.com/snowplow/snowplow/pull/3646/commits/d27455ecb1683c0913485748fe12e9ea1861db68 ?

knservis:
@BenFradet I mean this https://github.com/snowplow/snowplow/pull/3646/commits/f1addb6f3ddba4807251dbee8b07f86c8ce300ad#diff-35cd2b577f491cb5aeaf46ba8cead04eR23

BenFradet:
@knservis, yeah don't worry about this, this will be overhauled in #3624

knservis:
@BenFradet @chuwy This is now under review. Can you please have a look?

BenFradet:
will look at it tomorrow morning :+1:

3-enrich/stream-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/sources/AbstractSource.scala
```diff
@@ -154,15 +157,20 @@ abstract class AbstractSource(
 
   implicit val resolver: Resolver = igluResolver
 
+  def getPiiEvent(event: EnrichedEvent): Option[String] = {
+    if (event.pii == null || event.pii.isEmpty) None
+    else Some(event.pii)
+  }
```
BenFradet: isn't this just `Option(event.pii).filter(_.nonEmpty)` ?

3-enrich/stream-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/good/StructEventWithContextSpec.scala
```diff
@@ -174,7 +175,7 @@ class StructEventWithContextSpec extends Specification {
       // "-1" prevents empty strings from being discarded from the end of the array
       val fields = enrichedEvent.toOption.get._1.split("\t", -1)
       fields.size must beEqualTo(StructEventWithContextSpec.expected.size)
-
+      enrichedEvent.toOption.get._3.get must_== StructEventWithContextSpec.pii
```
BenFradet: can we add additional `isDefined` checks so that if it fails we'll have an idea why?

3-enrich/stream-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/PiiEmitSpec.scala
```diff
@@ -0,0 +1,197 @@
...
+          props.put("value.deserializer", "org.apache.kafka.common.serialization.StringDeserializer")
+          props.put("group.id", s"consumer-$topic")
+          props
+        }
+        val testConsumerPii = new KafkaConsumer[String, String](testKafkaPropertiesConsumer)
+        testConsumerPii.subscribe(List(topic))
+        var records = getRecords(testConsumerPii.poll(100))
+        while (((System.currentTimeMillis - started) / 1000 < timeoutSec - 1) && records.size < expectedRecords) {
+          records = records ++ getRecords(testConsumerPii.poll(100))
+        }
```
BenFradet: can't we just poll once for `timeoutSec`, also why do we need a future, we can just have this be synchronous it's tests?

3-enrich/stream-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/EnrichApp.scala
```diff
@@ -76,44 +79,57 @@ object EnrichApp {
...
+          else failure(s"Configuration file $f does not exist"))
+      opt[String]("resolver")
+        .required()
+        .valueName("<resolver uri>")
         .text(s"Iglu resolver file, $regexMsg")
         .action((r: String, c: FileConfig) => c.copy(resolver = r))
         .validate(_ match {
           case FilepathRegex(_) | DynamoDBRegex(_, _, _) => success
-          case _ => failure(s"Resolver doesn't match accepted uris: $regexMsg")
+          case _                                         => failure(s"Resolver doesn't match accepted uris: $regexMsg")
```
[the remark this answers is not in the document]
knservis: It looks like it does.. Hey alignment is more important right? 😄

3-enrich/stream-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/PiiEmitSpec.scala
```diff
@@ -0,0 +1,215 @@
...
+        records
+      }
```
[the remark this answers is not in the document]
knservis: Are you referring to the properties above? (`key.deserializer` etc.)

3-enrich/stream-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/SpecHelpers.scala
```diff
@@ -29,19 +27,32 @@ import common.utils.JsonUtils
...
 import model._
+import scala.util.matching.Regex
 import sources.TestSource
 
+/**
+ * Helper trait to get list of expected results that contain either a String or a Regex
+ */
+sealed trait StringOrRegex
+case class JustString(s: String) extends StringOrRegex
+case class JustRegex(r: Regex)   extends StringOrRegex
```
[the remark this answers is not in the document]
knservis: I'll try it out.

3-enrich/stream-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/PiiEmitSpec.scala
knservis: Yes we could, but the test would take a lot longer.

3-enrich/stream-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/good/StructEventWithContextSpec.scala
knservis: 👍

3-enrich/stream-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/sources/AbstractSource.scala
knservis: I think nonEmpty will not do the right thing is if the field is actually null

chuwy: `Option(null) == None`

knservis: 👍

3-enrich/stream-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/PiiEmitSpec.scala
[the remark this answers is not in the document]
BenFradet: I mean `producer`, `getConsumer`, `getRecords`, etc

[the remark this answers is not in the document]
knservis: As I don't know of the use case you have in mind (another test?) can you please tell me where exactly (package? SpecHelpers?) they need be extracted? Alternatively if you have an additional use case, I can refactor and add the new test to this release (lets create a ticket in this milestone for that). I think using tests like that more could be great as they will improve coverage.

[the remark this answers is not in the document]
BenFradet: >I think using tests like that more could be great as they will improve coverage.

That's exactly what I meant. So what would be easiest would be to extract all this into a trait so that a spec can just extend it and have kafka integration.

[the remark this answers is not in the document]
knservis: 👍

knservis:
@BenFradet Can you please have another look?

3-enrich/stream-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/EnrichApp.scala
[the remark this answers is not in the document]
BenFradet: could we limit the vertical alignments to circumvent this issue?

3-enrich/stream-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/sources/AbstractSource.scala
```diff
@@ -211,36 +234,48 @@ abstract class AbstractSource(
...
 
-    val successesTriggeredFlush = sink.get.map(_.storeEnrichedEvents(smallEnoughSuccesses))
-    val failuresTriggeredFlush = badSink.get.map(_.storeEnrichedEvents(failures ++ sizeBasedFailures))
-    if (successesTriggeredFlush == Some(true) || failuresTriggeredFlush == Some(true)) {
+    val anonymizedSuccesses = smallEnoughSuccesses.map({
+      case (event: String, partition: String, _) => (event, partition)
+    })
+    val piiSuccesses = smallEnoughSuccesses.flatMap({
+      case (_, partition: String, pii: Option[String]) => pii.map((_, partition))
+    })
```
BenFradet: unecessary parens here and above

3-enrich/stream-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/EnrichApp.scala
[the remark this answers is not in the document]
knservis: This is easy to workaround.

3-enrich/stream-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/sources/AbstractSource.scala
```diff
@@ -211,36 +234,48 @@ abstract class AbstractSource(
...
+    val successesTriggeredFlush = sink.get.map(_.storeEnrichedEvents(anonymizedSuccesses))
+    val piiTriggeredFlush       = piiSink.get.map(_.storeEnrichedEvents(piiSuccesses))
+    val failuresTriggeredFlush =
+      badSink.get.map(_.storeEnrichedEvents(failures ++ sizeBasedFailures))
+    if (successesTriggeredFlush == Some(true) || failuresTriggeredFlush == Some(true) || piiTriggeredFlush == Some(
+        true)) {
```
BenFradet: weird formatting :'(

3-enrich/stream-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/KafkaIntegrationSpec.scala
```diff
@@ -0,0 +1,125 @@
...
+import model.EnrichConfig
+import scalatracker.Tracker
+import iglu.client.Resolver
+import common.enrichments.EnrichmentRegistry
+
+/*
+ * Extending this trait creates a new integration test with a new instance of kafka
+ * See PiiEmitSpec for an example of how to use it
+ */
+
```
BenFradet: empty line

3-enrich/stream-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/KafkaIntegrationSpec.scala
```diff
@@ -0,0 +1,125 @@
...
+    resolver: Resolver,
+    registry: EnrichmentRegistry,
+    tracker: Option[Tracker]) = Future {
+    EnrichApp.run(configuration, resolver, registry, tracker)
+  }
+
+  def producerTimeoutSec: Int
+  def inputProduced = Try { Await.result(producer, Duration(s"$producerTimeoutSec sec")) }
+
+  def producer = Future {
```
BenFradet: shouldn't this be called `produce`?

BenFradet: you're missing the return type too, that'd be cool to make it a rule btw

3-enrich/stream-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/KafkaIntegrationSpec.scala
```diff
@@ -0,0 +1,125 @@
...
+    events.foreach { r =>
+      testProducer.send(new ProducerRecord(testGoodIn, "key", r))
+    }
+    testProducer.flush
+    testProducer.close
+  }
+  private def getRecords(cr: ConsumerRecords[String, String]): List[String] =
+    cr.map(_.value).toList
+
+  def getConsumer(topic: String, expectedRecords: Int, timeoutSec: Int) = Future {
```
BenFradet: shouldn't this be called `getRecords`, missing return type

3-enrich/stream-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/KafkaIntegrationSpec.scala
knservis: Fixed here. There is a rule for that I have added for that in scalastyle, so soon we will get the check.

3-enrich/stream-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/KafkaIntegrationSpec.scala
knservis: Fixed

3-enrich/stream-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/sources/AbstractSource.scala
knservis: Fixed

3-enrich/stream-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/sources/AbstractSource.scala
knservis: Fixed

3-enrich/stream-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/KafkaIntegrationSpec.scala
BenFradet: nice :+1:

3-enrich/stream-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/KafkaIntegrationSpec.scala
knservis: Changed

knservis:
@BenFradet Is there something else on this PR?

3-enrich/stream-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/KafkaIntegrationSpec.scala
```diff
@@ -0,0 +1,128 @@
...
+    getRecords(testPii, expectedPii, consumerExecutionTimeoutSec)
+  def allFutures: Future[(List[String], List[String], List[String])] =
+    for {
+      good <- producedGoodRecords
+      bad  <- producedBadRecords
+      pii  <- producedPiiRecords
+    } yield (good, bad, pii)
+  def outputResult: Try[(List[String], List[String], List[String])] = Try {
+    Await.result(allFutures, Duration(s"$consumerExecutionTimeoutSec sec"))
+  }
```
BenFradet: can't we do away with the `Try` / `Await.result` as it defeats the purpose of having futures in the first place?

we can maybe let specs2 do the matching directly on futures?

knservis: Sure. So that I understand the issue, can you please explain how it defeats the purpose of having futures in the first place?

BenFradet: You're building a future to just await its results synchronously so you lose the purpose of the future (async computations) and you wrap the result in a try, losing Failure / Success resulting of the future.

You might as well do without the future if you're doing synchronous computations wrapped in a try / catch.

Do you see what I mean?

knservis: I don't think I do understand.
Currently there are three parallel background tasks that will complete when they receive the number of results that they are expecting. As this may never happen there needs to be a timeout. I am sure you realise that the future returns before the timeout when the result is ready. If we were waiting for the three results synchronously we would need to wait for at least as long as the timeout (45 sec). Actually the consumers return in about 2 sec. 
So the objective here is to wait for a result that we don't know when it is coming (but usually pretty quickly in the successful case), but not wait forever, to do it in parallel for the three results expected so as not to delay the test unnecessarily. If the code in the future throws an exception, I see that exception in the Try. If the code times out it throws so the try is needed there. 
These results are needed to complete the test evaluation so they need to be awaited. 
I am happy to use the specs2 matcher (which looks almost identical in code):

```
  /**
   * when a Future contains a result, it can be awaited to return this result
   */
  implicit class futureAsResult[T : AsResult](f: Future[T]) {
    def await: Result = await()
    def await(retries: Int = 0, timeout: FiniteDuration = 1.seconds): Result = {
      def awaitFor(retries: Int, totalDuration: FiniteDuration = 0.seconds): Result = {
        try Await.result(f.map(value => AsResult(value)), timeout)
        catch {
          case e: TimeoutException =>
            if (retries <= 0) Failure(s"Timeout after ${totalDuration + timeout}")
            else awaitFor(retries - 1, totalDuration + timeout)
          case other: Throwable    => throw other
        }
      }
      awaitFor(retries)
    }
  }
```

however since I still don't understand what you mean by "lose the purpose of the future" I don't know if it will address your concern. I think this will be easier over HipChat..

knservis: Recapping @BenFradet we will use specs2 Matcher.await instead of Await.result to give the option to not block to whoever is implementing another kafka test in the future.

BenFradet: yes since specs2 doesn't do async testing apparently...

knservis:
@BenFradet please have another look.

3-enrich/stream-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/sources/AbstractSource.scala
```diff
@@ -237,13 +235,20 @@ abstract class AbstractSource(
       m <- MaxRecordSize
     } yield AbstractSource.oversizedSuccessToFailure(value, m) -> key
 
-
-    val anonymizedSuccesses = smallEnoughSuccesses.map({case (event: String, partition: String, _) => (event, partition)})
-    val piiSuccesses = smallEnoughSuccesses.flatMap({case (_, partition: String, pii: Option[String]) => pii.map((_, partition))})
+    val anonymizedSuccesses = smallEnoughSuccesses.map {
+      case (event: String, partition: String, _) => (event, partition)
```
chuwy: Type annotations should be unnecessary.

3-enrich/stream-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/sources/AbstractSource.scala
```diff
@@ -237,13 +235,20 @@ abstract class AbstractSource(
...
+    }
+    val piiSuccesses = smallEnoughSuccesses.flatMap {
+      case (_, partition: String, pii: Option[String]) => pii.map((_, partition))
```
chuwy: Also, type annotations should be unnecessary.

3-enrich/stream-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/sources/AbstractSource.scala
```diff
@@ -237,13 +235,20 @@ abstract class AbstractSource(
...
+    }
     val successesTriggeredFlush = sink.get.map(_.storeEnrichedEvents(anonymizedSuccesses))
-    val piiTriggeredFlush = piiSink.get.map(_.storeEnrichedEvents(piiSuccesses))
-    val failuresTriggeredFlush = badSink.get.map(_.storeEnrichedEvents(failures ++ sizeBasedFailures))
-    if (successesTriggeredFlush == Some(true) || failuresTriggeredFlush == Some(true) || piiTriggeredFlush == Some(true)) {
+
+    val piiTriggeredFlush       = piiSink.get.map(_.storeEnrichedEvents(piiSuccesses))
+    val failuresTriggeredFlush =
+      badSink.get.map(_.storeEnrichedEvents(failures ++ sizeBasedFailures))
+    if (successesTriggeredFlush == Some(true) ||
```
chuwy: Linter inside my IDE usually proposes to change these to `a.contains(true) || b.contains(true)`. Just a nit, not requirement.

3-enrich/stream-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/sources/AbstractSource.scala
```diff
@@ -154,15 +157,20 @@ abstract class AbstractSource(
 
   implicit val resolver: Resolver = igluResolver
 
+  def getPiiEvent(event: EnrichedEvent): Option[String] = {
+    if (event.pii == null || event.pii.isEmpty) None
```
chuwy: So, it wasn't changed in the end?

knservis: This was changed: https://github.com/snowplow/snowplow/pull/3646/files#diff-e405d498f6ff411d9e07f91778055e48R173 I think you are looking at an outdated version..

knservis:
@alexanderdean @BenFradet The PII is not an enriched event. Can you please check that the fields (in the table above make sense)?

BenFradet:
- `platform` is user-defined, I don't know if we'd want to change that
- should we change the timestamps ? (`elt_tstamp` and `derived_tstamp`)
- shouldn't the pii transformations be split (e.g. one event per pii transformation, I remember reading `List[EnrichedEvent]`)?

I'll let Alex as TPM handle those qs

knservis:
It seems like we will also need a new event to be included in the contexts field that will contain the parent event id. The format will be this https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/duplicate/jsonschema/1-0-0 (except for the name and possibly the field name)

knservis:
@BenFradet I have finished rebasing and all some changes that we missed on PR (moving integration test to separate module and making piistream optional). Can you please have another look to make sure I did not miss anything?

3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/sources/Source.scala
```diff
@@ -158,35 +209,46 @@ abstract class Source(
...
-    if (successesTriggeredFlush == true || failuresTriggeredFlush == true) {
+    val anonymizedSuccesses = smallEnoughSuccesses.map {
+      case (event, partition, _) => (event, partition)
+    }
+    val piiSuccesses = smallEnoughSuccesses.flatMap {
+      case (_, partition, pii) => pii.map((_, partition))
+    }
+
+    val successesTriggeredFlush = goodSink.get.storeEnrichedEvents(anonymizedSuccesses)
+    val piiTriggeredFlush       = piiSink.map(_.get.storeEnrichedEvents(piiSuccesses)).getOrElse(false)
```
BenFradet: one thing I missed during the 1st review is, should we check for the sizes of the pii events separately from the usual enriched events?

3-enrich/stream-enrich/core/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/SpecHelpers.scala
```diff
@@ -141,123 +150,176 @@ object SpecHelpers {
...
+  }
+  """
 
-    implicit val resolver: Resolver = validatedResolver.fold(
-      e => throw new RuntimeException(e),
-      s => s
-    )
+  val igluConfig = {
+    val resolverEnvVar = for {
+      config <- sys.env.get("ENRICH_RESOLVER_CONFIG")
```
BenFradet: why is this needed? it should either be all env vars or no env vars?

3-enrich/stream-enrich/examples/config.hocon.sample
```diff
@@ -51,28 +53,29 @@ enrich {
       # 'kafka' for reading / writing to a Kafka topic
       # 'nsq' for reading / writing to a Nsq topic
       # 'stdin' for reading from stdin and writing to stdout and stderr
-      enabled = kinesis
+      type =  {{ sinkType }}
```
BenFradet: please modify the comment above "Configuration shown..."

3-enrich/stream-enrich/integration/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/KafkaIntegrationSpec.scala
```diff
@@ -0,0 +1,143 @@
...
+ * See PiiEmitSpec for an example of how to use it
+ */
+trait KafkaIntegrationSpec
+    extends EmbeddedKafkaContext
+    with DefaultKafkaPorts
+    with TryMatchers
+    with TraversableMatchers {
+
+  import KafkaIntegrationSpecValues._
+  implicit val ec = ExecutionContext.fromExecutor(new ForkJoinPool(16))
```
BenFradet: can't we use the default one

3-enrich/stream-enrich/integration/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/KafkaIntegrationSpec.scala
```diff
@@ -0,0 +1,143 @@
...
+  def getMainApplicationFuture(
+                                configuration: StreamsConfig,
+                                resolver: Resolver,
+                                registry: EnrichmentRegistry,
+                                tracker: Option[Tracker]): Future[Unit] = Future {
+    KafkaEnrich.getSource(configuration, resolver, registry, tracker).toOption.get.run()
+  }
+
+  def producerTimeoutSec: Int
+  def inputProduced: Try[Unit] = Try { Await.result(produce, Duration(s"$producerTimeoutSec sec")) }
```
BenFradet: why is future used?

3-enrich/stream-enrich/integration/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/KafkaIntegrationSpec.scala
```diff
@@ -0,0 +1,143 @@
...
+  def produce: Future[Unit] = Future {
+    val testKafkaPropertiesProducer = {
+      val props = new Properties()
+      props.put("bootstrap.servers", kafkaHost)
+      props.put("client.id", "producer-george")
+      props.put("key.serializer", "org.apache.kafka.common.serialization.StringSerializer")
+      props.put("value.serializer", "org.apache.kafka.common.serialization.ByteArraySerializer")
+      props
+    }
```
BenFradet: can't those properties be extracted, no need to recreate one every time we call produce

3-enrich/stream-enrich/integration/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/KafkaIntegrationSpec.scala
```diff
@@ -0,0 +1,143 @@
...
+
+    val testProducer = new KafkaProducer[String, Array[Byte]](testKafkaPropertiesProducer)
+    val events = inputGood
+    events.foreach { r =>
+      testProducer.send(new ProducerRecord(testGoodIn, "key", r))
```
BenFradet: why are you using a future if you're discarding the result of the only async code here?

3-enrich/stream-enrich/integration/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/PiiEmitSpec.scala
```diff
@@ -0,0 +1,159 @@
...
+
+// This project
+import good._
+import model.StreamsConfig
+
+class PiiEmitSpec extends Specification with FutureMatchers {
+
+  import KafkaIntegrationSpecValues._
+
+
```
BenFradet: empty line

3-enrich/stream-enrich/build.sbt
```diff
@@ -100,3 +102,15 @@ lazy val stdin = project
   .settings(moduleName := "snowplow-stream-enrich-stdin")
   .settings(allSettings)
   .dependsOn(core)
+
+lazy val integration = project.in(file("./integration"))
```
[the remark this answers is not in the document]
knservis: ok

3-enrich/stream-enrich/build.sbt
```diff
@@ -100,3 +102,15 @@ lazy val stdin = project
...
+  .settings(moduleName := "integration")
+  .settings(allSettings)
+  .settings(BuildSettings.addExampleConfToTestCp)
+  .settings(libraryDependencies ++= Seq(
+    Dependencies.Libraries.kafkaClients,
+        // Test
```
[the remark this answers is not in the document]
knservis: ok

3-enrich/stream-enrich/integration/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/KafkaIntegrationSpec.scala
knservis: There are futures that have only some effect and don't return anything..

3-enrich/stream-enrich/integration/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/PiiEmitSpec.scala
knservis: ok

3-enrich/stream-enrich/integration/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/KafkaIntegrationSpec.scala
knservis: sure

3-enrich/stream-enrich/core/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/SpecHelpers.scala
knservis: This how it is done in scala-common-enrich (it uses an env var to override the default iglu repo). I am not sure how it should be done otherwise (I don't understand all/no env vars).

3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/sources/Source.scala
knservis: I am assuming that you are referring to the lines below `if(successesTriggeredFlush == true...`. Is that correct?

alexanderdean:
Hey @knservis - huge apologies for the delay in getting to:

>Looks like the emitted PII event should be an actual EnrichedEvent here is how I propose to populate it (sorry about the wide format) @alexanderdean

It mostly looks good, but I suggest some changes:

* `stream-enrich` is an invalid `platform` value, let's use `srv`
* Let's use same (not new) as `etl_tstamp`
* Let's not set the dvce_created_tstamp
* Let's set the true_tstamp to the same as the derived_tstamp (i.e. a new timestamp)

That table also shouldn't have the pii column at the bottom - that pii column is a hack in the code, it's not a formal part of the enriched event.

Thanks!

3-enrich/stream-enrich/core/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/SpecHelpers.scala
BenFradet: yes let's leave it as is ... as discussed it can be useful when testing and using a specific registry without having to modify the code itself

3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/sources/Source.scala
knservis: Ok I understood now. So we can definitely check pii for size, but the question is, what to do if the are too big (they are still PII)?

3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/sources/Source.scala
```diff
@@ -101,37 +112,74 @@ abstract class Source(
...
+  private val formatter           = DateTimeFormat.forPattern("yyyy-MM-dd HH:mm:ss.SSS")
+  private val PII_EVENT_NAME      = "pii_transformation"
+  private val PII_EVENT_VENDOR    = "com.snowplowanalytics.snowplow"
+  private val PII_EVENT_FORMAT    = "jsonschema"
+  private val PII_EVENT_VERSION   = "1-0-0"
+  private val PII_EVENT_PLATFORM  = "stream-enrich"
+  def getPiiEvent(event: EnrichedEvent): Option[EnrichedEvent] =
+    Option(event.pii)
+      .filter(_.nonEmpty)
+      .map(piiStr => {
```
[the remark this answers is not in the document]
knservis: ok

3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/sources/Source.scala
knservis: @alexanderdean Any suggestions?

alexanderdean: What are our options here @knservis ?

knservis: I can see the following options:

* Do nothing: As this is using the *small enough* Enriched Events to emit PII we can assume that the PII event is going to be smaller than the parent event. This is a reasonable assumption IMHO. 
* Use the same limit and test PII separately which then creates the following two options for oversize PII events: 
* * If PII are bigger than the limit, throw them away (will create data gaps in piinguin)
* * Send them to the existing *bad* stream. That is problematic as it would leak PII to the bad stream, unless we somehow scrub the original values. Sounds like a much hassle for no actual benefit as there would still need to be a lot of manual work to recover.
* * Send the parent event _id and an appropriate message to *bad* (but if we are doing that, why not send it to PII and have alternative handling where we look up the parent event and do PII backfilling).
* Come up with an EnrichedEvent fragmentation protocol to handle all cases where the event is too large, where they are fragmented and reassembled. This may seem like a good option and it certainly will avoid dropping data, but it is problematic in that most (all?) projects operate on single events at the moment.

alexanderdean: Thanks for the detailed options @knservis ! Let's go with do nothing, but:

1. Create a ticket to flag that we are doing nothing, so we don't forget about this. Heavily link this new ticket back to this thread
2. Create a placeholder for exploring the fragmentation protocol idea, which is really neat. Again, heavily link it back to here

Thanks! @BenFradet shout if you disagree with the plan.

knservis: Memo: https://github.com/snowplow/snowplow/issues/3755

Proposal: https://github.com/snowplow/snowplow/issues/3756
👍

knservis:
OK @alexanderdean and @BenFradet are you ok with this PR?

BenFradet:
@knservis ping me when the tests and commits are sorted

knservis:
@BenFradet tests are sorted. Not sure what you mean about commits

3-enrich/stream-enrich/pubsub/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/sources/GooglePubSubSource.scala
```diff
@@ -58,36 +58,47 @@ object GooglePubSubSource {
...
+      else
+        None.success
     badPublisher <- GooglePubSubSink
       .validateAndCreatePublisher(googlePubSubConfig, config.buffer, config.out.bad)
       .validation
     topic = ProjectTopicName.of(googlePubSubConfig.googleProjectId, config.in.raw)
     subName = ProjectSubscriptionName.of(googlePubSubConfig.googleProjectId, config.appName)
     _ <- toEither(createSubscriptionIfNotExist(subName, topic)).validation
-  } yield new GooglePubSubSource(goodPublisher, badPublisher, igluResolver,
+  } yield new GooglePubSubSource(goodPublisher,piiPublisher,  badPublisher, igluResolver,
```
BenFradet: space

PR #3669, opened 2018-03-08
Added salt and relevant changes

knservis:


3-enrich/scala-common-enrich/project/Dependencies.scala
```diff
@@ -101,6 +103,8 @@ object Dependencies {
     val scalaWeather     = "com.snowplowanalytics"      %% "scala-weather"             % V.scalaWeather
     val scalaj           = "org.scalaj"                 %% "scalaj-http"               % V.scalaj
     val gatlingJsonpath  = "io.gatling"                 %% "jsonpath"                  % V.gatlingJsonpath
+    // Java
+    val ssm              = "com.amazonaws"              % "aws-java-sdk-ssm"           % V.aws
```
BenFradet: mmmh, sce can't really depend on aws unfortunately.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/pii/PiiPseudonymizerEnrichment.scala
```diff
@@ -0,0 +1,311 @@
...
+  }.leftMap(_.toProcessingMessageNel)
+
+  private[pii] def getHashFunction(strategyFunction: String): Validation[String, DigestFunction] =
+    strategyFunction match {
+      case "MD2"     => { DigestUtils.md2Hex(_: Array[Byte]) }.success
+      case "MD5"     => { DigestUtils.md5Hex(_: Array[Byte]) }.success
+      case "SHA-1"   => { DigestUtils.sha1Hex(_: Array[Byte]) }.success
+      case "SHA-256" => { DigestUtils.sha256Hex(_: Array[Byte]) }.success
+      case "SHA-384" => { DigestUtils.sha384Hex(_: Array[Byte]) }.success
+      case "SHA-512" => { DigestUtils.sha512Hex(_: Array[Byte]) }.success
```
BenFradet: can't we just have those lines as: `DigestUtils.sha512Hex(_: Array[Byte]).success`?

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/pii/PiiPseudonymizerEnrichment.scala
```diff
@@ -0,0 +1,311 @@
...
+
+  /**
+   * Tests whether the schema for this event matches the schema criterion and if it does modifies it.
+   */
+  private def modifyObjectIfSchemaMatches(context: List[(String, json4s.JValue)],
+                                          strategy: PiiStrategy): (JObject, List[JsonModifiedField]) = {
+    val fieldsObj = context.toMap
+    (for {
+      schema <- fieldsObj.get("schema")
+      schemaStr = schema.extract[String]
```
BenFradet: can't this fail?

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/pii/PiiPseudonymizerEnrichment.scala
```diff
@@ -0,0 +1,311 @@
...
+      (compacted, modifiedFields.toList)
+    } else (null, modifiedFields.toList)
+  }
+
+  /**
+   * Modifies the object if applicable and adds to the list of modified values, if applicable.
+   */
+  private def modifyAndGetValues(context: List[(String, json4s.JValue)],
+                                 strategy: PiiStrategy,
+                                 modifiedFields: MutableList[JsonModifiedField]): JObject = {
```
BenFradet: can't this function return `listOfModifiedValues` instead of passing and mutating `modifiedFields`?

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/pii/package.scala
```diff
@@ -0,0 +1,120 @@
...
+  // SUPPRESS_EXCEPTIONS is useful here as we prefer an empty list to an exception when a path is not found.
+  private[pii] val JsonPathConf =
+    Configuration
+      .builder()
+      .options(JOption.SUPPRESS_EXCEPTIONS)
+      .jsonProvider(new JacksonJsonNodeJsonProvider(JacksonNodeJsonObjectMapper))
+      .build()
+}
+
+package pii {
```
BenFradet: why can't we move everything into the `pii` object? package object are ideally to avoid since they'll be removed

alexanderdean:
I wanted to dig into:

> aws dependency in sce, I think this is a no-go unfortunately

I can probably guess, but can you set out why, @BenFradet ? If we can't do this, then we will probably have to wait on the Snowplow Control Plane before we can probably secure secrets that are otherwise embedded directly into Snowplow enrichment configs...

BenFradet:
The whole recent effort on enrich and ssc was purposely not to embed aws dependencies if you're not on aws. If these get into the sce we're back to square one.

What we could do is decline the sce into sce and sce-aws and have aws specific artifacts depend on it.

Of course this doesn't make this specific part of the enrichment available to people on prem or on gcp but that was already the case...

alexanderdean:
> What we could do is decline the sce into sce and sce-aws and have aws specific artifacts depend on it.

I think it's better if we just park the AWSisms and use this as another reason to push on with Control Plane.

@knservis are you okay to back the AWS bits out into a parked branch?

knservis:
@alexanderdean Sure. In fact I believe it is better that way. As an aside would it not be better to have those as environment variables anyway?

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/pii/PiiPseudonymizerEnrichment.scala
knservis: No that doesn't compile

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/pii/PiiPseudonymizerEnrichment.scala
knservis: Done (good point although readability is worse now if you ask me)

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/pii/PiiPseudonymizerEnrichment.scala
knservis: You mean if the schema is not a string? I suppose its conceivable. Fixed anyway

knservis:
@BenFradet I am keen to tidy up all the branches on this release. Is the PR good enough to merge (after rebase)?

3-enrich/scala-common-enrich/project/Dependencies.scala
```diff
@@ -61,6 +61,8 @@ object Dependencies {
     val scalaj           = "2.3.0"
     val gatlingJsonpath  = "0.6.4"
     val scalaUri         = "0.5.0"
+    // Java
+    val aws              = "1.11.290"
```
BenFradet: I thought this had been parked into its own branch?

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/pii/PiiPseudonymizerEnrichment.scala
```diff
@@ -0,0 +1,283 @@
...
+              case (k: String, x: JValue) => (k, (x, List.empty[JsonModifiedField]))
+            }
+          (JObject(contextMapped.mapValues(_._1).toList), contextMapped.values.map(_._2).flatten)
+        }
+        case x => (x, List.empty[JsonModifiedField])
+      }
+      val rendered  = render(parsedAndSubistuted)
+      val compacted = compact(rendered)
+      (compacted, modifiedFields)
+    } else (null, List.empty[JsonModifiedField])
```
BenFradet: can we split this function into smaller chunks?

knservis:
@BenFradet Thank you for the review. Can you see anything else?

alexanderdean:
> As an aside would it not be better to have those as environment variables anyway?

Env variables don't deploy particularly well (the seed value for the environment variable itself has to live somewhere).

Issue #3556, opened 2017-12-29
Scala Stream Collector: move cross domain configuration to enabled-style


BenFradet:
At the moment, there are configurations which are defined as:

```hocon
x {
  enabled = true/false
  restOfTheConfig = ...
}
```

and others defined as:

```hocon
x {
  restOfTheConfig = ...
}
```

and its presence or absence determines if the feature is used.

We should pick one and stick to it.

I tend to prefer the second one as less stuff you don't care about gets in your way.

wdyt @chuwy @alexanderdean ?

alexanderdean:
Hmm good point. I think I prefer the first one because it's more templating friendly.  @chuwy what do you prefer?

chuwy:
At least in JSON configurations we prefer first approach (even have linter for that). I also think that first approach is more explicit.

BenFradet:
from @rbolkey :

>I looked at the thread there. I didn't see anyone propose using both Option at the top level along with an enabled field? You can kind of have both pieces of cake there. Missing block implies disabled (less pain for user's configuring). Present block has enabled defaulting to true, but you can template it to false if desired? Thoughts?

My personal view is that it might be kinda confusing, wdyt @chuwy @alexanderdean ?

alexanderdean:
Not sure I quite understand @rbolkey's suggestion?

BenFradet:
both

```conf
a {
  x {
    enabled = true/false
    restOfTheConfig = ...
  }
}
```

```conf
a {
}
```

are valid

alexanderdean:
Not sure. Your call.

PR #3670, opened 2018-03-08
Scala Stream Collector: add a way to customize the response from the root path

rbolkey:
We needed to have the ability to customize the response from the root path ("/") of the streaming collector. For our particular case, we need to allow the collector to redirect to our website and not just report "404 not found".

This is similar to the behavior of other data collector endpoints like:

https://secure.adnx.com/
https://dpm.demdex.net/

I've kept the default behavior as the current behavior.

Does this functionality look acceptable?


alexanderdean:
This is actually really cool @rbolkey.

We were thinking of adding something similar for a totally different reason: letting Snowplow operators provide information about *what this collector is* on the root to help data subjects understand who is collecting their data.

rbolkey:
Great. I had that idea in the back of my mind as well. I've been wading through GDPR (as have you all) over the past month.

How close is functionality to what you all were envisioning? What's missing or should be different?

alexanderdean:
It looks good to me - but let me add @yalisassoon as he may have some thoughts...

snowplowcla:
@rbolkey has signed the Software Grant and Corporate Contributor License Agreement

2-collectors/scala-stream-collector/examples/config.hocon.sample
```diff
@@ -76,6 +76,19 @@ collector {
...
+  # Useful if you need to redirect to web content or privacy policies regarding the use of this collector.
+  rootResponse {
+    # Optional, defaults to 404
+    statusCode = 302
+    headers = {
+        Location = "https://127.0.0.1/",
+        X-Custom = "something"
+    }
+    # Optional, defaults to empty string
+    body = "302, redirecting"
```
alexanderdean: Do we need a content type?

yalisassoon:
I love this! From a GDPR perspective it'd be an awesome place to make clear:

(i) who is collecting the data
(ii) on what basis data is collected
(iii) how data subjects can exercise their rights

on the page in question.

Massive thanks @rbolkey !

2-collectors/scala-stream-collector/examples/config.hocon.sample
rbolkey: I was curious if the entire HttpEntity could be stored in HOCON, but didn't push it that far.

BenFradet:
Thanks @rbolkey , I'll have a look on Monday :+1:

2-collectors/scala-stream-collector/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala
```diff
@@ -117,6 +122,7 @@ package model {
     cookie: CookieConfig,
     cookieBounce: CookieBounceConfig,
     redirectMacro: RedirectMacroConfig,
+    rootResponse: Option[RootResponseConfig],
```
BenFradet: we've standardized against enabled-style in #3556, it'd be cool if this new feature could follow this guildeline too :+1:

2-collectors/scala-stream-collector/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala
```diff
@@ -61,6 +61,11 @@ package model {
     enabled: Boolean,
     placeholder: Option[String]
   )
+  final case class RootResponseConfig(
+    statusCode: Option[Int] = None,
+    headers: Option[Map[String, String]] = None,
+    body: Option[String] = None
```
BenFradet: I think we can have `404` as default for `statusCode` and `Map.empty[String, String]` for `headers`, wdyt?

2-collectors/scala-stream-collector/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
```diff
@@ -157,6 +158,26 @@ class CollectorService(
...
+  def rootResponse(unknownRoute: Option[RootResponseConfig]): HttpResponse =
+    unknownRoute match {
+      case Some(c) =>
+        val statusCode = c.statusCode.getOrElse(404)
+        HttpResponse(
+          statusCode,
+          c.headers.map((headersMap: Map[String, String]) =>
+            headersMap
+              .map((tuple: (String, String)) => RawHeader(tuple._1, tuple._2))
+              .to[collection.immutable.Seq]
```
BenFradet: you should be able to do:

```scala
headersMap.map { case (k, v) => RawHeader(k, v) }.toSeq
```

2-collectors/scala-stream-collector/examples/config.hocon.sample
BenFradet: can't the content-type be specified through the headers anyway?

alexanderdean:
> can't the content-type be specified through the headers anyway?

Oh, good point

2-collectors/scala-stream-collector/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
rbolkey: Thanks for the pointer. I think I can switch to the case statement. The `toSeq` seems to have troubles due to its return type. It returns `scala.collection.Seq`, but `HttpResponse` is expecting a more specific type, `scala.collection.immutable.Seq`.

2-collectors/scala-stream-collector/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala
rbolkey: Will do. Ironically, that's what my first pass at this code did :-).

2-collectors/scala-stream-collector/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala
rbolkey: No issues with that.

rbolkey:
@BenFradet applied changes based on your suggestions.

knservis:
@rbolkey Can you please change the base branch of the PR to release/gdpr-2 and rebase?

rbolkey:
Not up to speed on the travis checks. Tests pass locally :-).

knservis:
@rbolkey Thanks for that. Yes there have been issues with travis. @BenFradet is that related with issues you had with travis?

2-collectors/scala-stream-collector/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
```diff
@@ -157,6 +160,18 @@ class CollectorService(
       case None => HttpResponse(404, entity = "404 not found")
     }
 
+
+  override def rootResponse: HttpResponse =
+    rootResponse(config.rootResponse)
+
+  def rootResponse(c: RootResponseConfig): HttpResponse =
+    if (c.enabled) {
+      val rawHeaders = if (c.headers.nonEmpty) c.headers.map { case (k, v) => RawHeader(k, v) }.toSeq.to[immutable.Seq] else Nil
```
BenFradet: why is `to[immutable.Seq]` needed?

BenFradet: no need for the `if (c.headers.nonEmpty)` btw

2-collectors/scala-stream-collector/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala
```diff
@@ -61,6 +61,12 @@ package model {
     enabled: Boolean,
     placeholder: Option[String]
   )
+  final case class RootResponseConfig(
+    enabled: Boolean = false,
+    statusCode: Int = 404,
+    headers: Map[String, String] = Map.empty[String, String],
+    body: String = ""
```
BenFradet: on second thought, I don't think we need default values since config parsing will make sure everything is in order

2-collectors/scala-stream-collector/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
rbolkey: @BenFradet I accidentally kept the `toSeq` in there, but the `to[immutable.Seq]` is necessary unless there is an implicit that I should add instead? Without the `to[immutable.Seq]` I get the following compilation error:

```
[error] /Users/rbolkey/src/onespot/snowplow/2-collectors/scala-stream-collector/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala:170: type mismatch;
[error]  found   : Seq[akka.http.scaladsl.model.headers.RawHeader]
[error]  required: scala.collection.immutable.Seq[akka.http.scaladsl.model.HttpHeader]
[error]       HttpResponse(c.statusCode, rawHeaders, HttpEntity(c.body))
```

rbolkey: I can drop the if check. The default for the headers parameter to `HttpResponse` is `Nil`, so I wasn't sure if it was wasteful to pass an empty map.

BenFradet: mmh weird, I'll have a look

2-collectors/scala-stream-collector/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala
rbolkey: I may need some more context on desired behavior here. If functionality it truly optional, should a user need to put any configuration at all in the config file?

As it stands in this commit, the `rootResponse` key will need to be defined in the configuration file (when the value was an `Option`) it did not need to be added, and the config was backward compatible. 

If I start removing the default values, a user will need to define all of them in their configuration file even if they don't want to use this feature?

Current behavior if `rootResponse` block left out of hocon file:
```
java -jar ./target/scala-2.11/snowplow-stream-collector-0.12.0.jar --config ./examples/config.hocon
Exception in thread "main" pureconfig.error.ConfigReaderException: Cannot convert configuration to a com.snowplowanalytics.snowplow.collectors.scalastream.model.CollectorConfig. Failures are:
  at 'rootResponse':
    - (file:/Users/rbolkey/src/onespot/snowplow/2-collectors/scala-stream-collector/./examples/config.hocon:20) Key not found.
```

Behavior if I have an empty `rootResponse` block but remove the `headers` default value from `model.scala` and I don't define it in hocon file:

```
java -jar ./target/scala-2.11/snowplow-stream-collector-0.12.0.jar --config ./examples/config.hocon
Exception in thread "main" pureconfig.error.ConfigReaderException: Cannot convert configuration to a com.snowplowanalytics.snowplow.collectors.scalastream.model.CollectorConfig. Failures are:
  at 'rootResponse.headers':
    - (file:/Users/rbolkey/src/onespot/snowplow/2-collectors/scala-stream-collector/./examples/config.hocon:81) Key not found.
```

rbolkey: As it stands, people do need to provide full configuration of the `cookieBounce`, `cookie`, `redirectMacro`, etc even if they don't want to use this feature, which is kind of a pain point.

BenFradet: yes, so there are two levels to this issue:

- we've opted out of maintaining backwards compat for the config in #3556 so `RootResponseConfig` can't be an option itself
- inside `RootResponseConfig`, we can have options but in this case I don't think having options / default values for everything makes sense? maybe for body and headers but not for statusCode, wdyt?

I'm generally against options for the config as this might result in hidden behaviours.

BenFradet: > As it stands, people do need to provide full configuration of the cookieBounce, cookie, redirectMacro, etc even if they don't want to use this feature, which is kind of a pain point.

I agree but it makes it harder to template as discussed in #3556

rbolkey: > inside RootResponseConfig, we can have options but in this case I don't think having options / default values for everything makes sense? maybe for body and headers but not for statusCode, wdyt?

That makes sense to me. It's not surprising for free-form Strings and Collections to default to the empty value. The status code is enum like and doesn't have a sensible default. Closest was following the previous behavior.

rbolkey: > I agree but it makes it harder to template as discussed in #3556

I looked at the thread there. I didn't see anyone propose using both Option at the top level along with an `enabled` field? You can kind of have both pieces of cake there. Missing block implies disabled (less pain for user's configuring). Present block has `enabled` defaulting to true, but you can template it to false if desired? Thoughts?

BenFradet: I think it's a bit too ambiguous to have both.

2-collectors/scala-stream-collector/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
BenFradet: Fixed it with:

```scala
val rawHeaders = c.headers.map { case (k, v) => RawHeader(k, v) }.toList
```

`Seq` are truly evil :(

2-collectors/scala-stream-collector/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala
rbolkey: That's fine. I will attest that it was an annoying nuisance to have to deal with when configuring and upgrading the collector, but I'm just one data point :-).

rbolkey:
@BenFradet Applied those updates.

rbolkey:
Also, not completely sure why I have so many changed files in this commit. Does it look like I botched a rebase or need to sync with the upstream?

2-collectors/scala-stream-collector/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
```diff
@@ -32,13 +32,16 @@ import model._
 import sinks.KinesisSink
 import utils.SplitBatch
 
+import scala.collection.immutable
```
BenFradet: no need for the import anymore

rbolkey:
Rebased again.

knservis:
@yalisassoon and @alexanderdean Is there missing?

BenFradet:
doesn't make sense to include this in R106 since it doesn't touch the collector

alexanderdean:
This can be added to a later release.

BenFradet:
incorporated into #3824 , closing. Thanks a lot :+1:

PR #3688, opened 2018-03-14
Adding Vero adapter to enrichment

misterpig:
Adapter for Vero webhook events. "user_updated" event is unique in that the "type" field is named "action". Only other conversion needed outside of the cleanupJsonEventValues method are for all the date-time values in the "triggered_at" field.
Schemas this will be validated against can be found [here](https://github.com/snowflake-analytics/iglu-central/tree/vero/schemas/com.getvero).

### DIFF
--- a/3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/AdapterRegistry.scala
+++ b/3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/AdapterRegistry.scala
@@ -53,7 +53,7 @@ object AdapterRegistry {
     val StatusGator     = "com.statusgator"
     val Unbounce        = "com.unbounce"
     val UrbanAirship    = "com.urbanairship.connect"
-    val Vero            = "com.vero"
+    val Vero            = "com.getvero"
   }
 
   /**
--- a/3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/AdapterRegistry.scala
+++ b/3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/AdapterRegistry.scala
@@ -53,6 +53,7 @@ object AdapterRegistry {
     val StatusGator     = "com.statusgator"
     val Unbounce        = "com.unbounce"
     val UrbanAirship    = "com.urbanairship.connect"
+    val Vero            = "com.vero"
   }
 
   /**
@@ -87,6 +88,7 @@ object AdapterRegistry {
       case (Vendor.StatusGator, "v1")           => StatusGatorAdapter.toRawEvents(payload)
       case (Vendor.Unbounce, "v1")              => UnbounceAdapter.toRawEvents(payload)
       case (Vendor.UrbanAirship, "v1")          => UrbanAirshipAdapter.toRawEvents(payload)
+      case (Vendor.Vero, "v1")                  => VeroAdapter.toRawEvents(payload)
       case _ =>
         s"Payload with vendor ${payload.api.vendor} and version ${payload.api.version} not supported by this version of Scala Common Enrich".failNel
     }
--- a/3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/VeroAdapter.scala
+++ b/3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/VeroAdapter.scala
@@ -71,8 +71,6 @@ object VeroAdapter extends Adapter {
   /**
    * Converts a payload into a single validated event
    * Expects a valid json returns failure if one is not present
-   * For "user_updated" events the field is named "action" instead of "type"
-   * in which case the event type will be set to Some("user_updated")
    *
    * @param json Payload body that is sent by Vero
    * @param payload The details of the payload
@@ -85,9 +83,8 @@ object VeroAdapter extends Adapter {
         case Success(p) => p.successNel
         case Failure(e) => s"$VendorName event failed to parse into JSON: [${e.getMessage}]".failureNel
       }
-      eventType = parsed.toOption.flatMap(p => (p \ "type").extractOpt[String]).getOrElse("user_updated")
-      formattedEvent = if (eventType == "user_updated") parsed
-      else cleanupJsonEventValues(parsed, ("type", eventType).some, s"${eventType}_at")
+      eventType = (parsed \ "type").extract[String]
+      formattedEvent   = cleanupJsonEventValues(parsed, ("type", eventType).some, s"${eventType}_at")
       reformattedEvent = reformatParameters(formattedEvent)
       schema <- lookupSchema(eventType.some, VendorName, EventSchemaMap)
       params = toUnstructEventParams(TrackerVersion, toMap(payload.querystring), schema, reformattedEvent, "srv")
@@ -122,7 +119,7 @@ object VeroAdapter extends Adapter {
 
   /**
    * Returns an updated Vero event JSON where
-   * the "action" field has been removed
+   * the "_tag" field is renamed to "tag"
    * and "triggered_at" fields' values have been converted
    *
    * @param json The event JSON which we need to
@@ -136,13 +133,9 @@ object VeroAdapter extends Adapter {
       JString(JsonSchemaDateTimeFormat.print(dt))
     }
 
-    json
-      .transformField {
-        case ("triggered_at", JInt(value)) => ("triggered_at", toStringField(value.toLong * 1000))
-      }
-      .removeField {
-        case ("action", _) => true
-        case _             => false
-      }
+    json transformField {
+      case ("_tags", JObject(v))         => ("tags", JObject(v))
+      case ("triggered_at", JInt(value)) => ("triggered_at", toStringField(value.toLong * 1000))
+    }
   }
 }
--- a/3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/VeroAdapter.scala
+++ b/3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/VeroAdapter.scala
@@ -36,7 +36,9 @@ import org.joda.time.format.{DateTimeFormat, DateTimeFormatter}
 
 // This project
 import com.snowplowanalytics.snowplow.enrich.common.loaders.CollectorPayload
-import com.snowplowanalytics.snowplow.enrich.common.utils.{JsonUtils => JU}
+//import com.snowplowanalytics.snowplow.enrich.common.utils.{JsonUtils => JU}
+
+import scala.util.{Failure, Success, Try}
 
 /**
  * Transforms a collector payload which conforms to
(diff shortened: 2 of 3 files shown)


alexanderdean:
Awesome work @misterpig !

snowplowcla:
@misterpig has signed the Software Grant and Corporate Contributor License Agreement

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/AdapterRegistry.scala
```diff
@@ -53,6 +53,7 @@ object AdapterRegistry {
     val StatusGator     = "com.statusgator"
     val Unbounce        = "com.unbounce"
     val UrbanAirship    = "com.urbanairship.connect"
+    val Vero            = "com.vero"
```
miike: Looks like we missed an instance here. `com.vero` => `com.getvero`

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/VeroAdapter.scala
```diff
@@ -0,0 +1,167 @@
...
+   * Expects a valid json returns failure if one is not present
+   * For "user_updated" events the field is named "action" instead of "type"
+   * in which case the event type will be set to Some("user_updated")
+   *
+   * @param body_json Payload body that is sent by Vero
+   * @param payload The details of the payload
+   * @return a Validation boxing either a NEL of RawEvents on
+   *         Success, or a NEL of Failure Strings
+   */
+  private def payloadBodyToEvent(body_json: String, payload: CollectorPayload): Validated[RawEvent] =
```
BenFradet: we use camel case

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/VeroAdapter.scala
```diff
@@ -0,0 +1,167 @@
...
+      val formattedEvent = eventType match {
+        case Some("user_updated") => reformatParameters(parsed)
+        case _ => {
+          cleanupJsonEventValues(
+            parsed,
+            ("type", eventType.get).some,
+            eventType.get.concat("_at")
+          )
+        }
+      }
```
BenFradet: this can be:

```scala
      val eventType = (parsed \ "type").extractOpt[String].getOrElse("user_updated")
      val formattedEvent =
        if (eventType == "user_updated") reformatParameters(parsed)
        else cleanupJsonEventValues(parsed, ("type", eventType).some, s"$eventType_at")
```

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/VeroAdapter.scala
```diff
@@ -0,0 +1,167 @@
...
+
+      val reformattedEvent = reformatParameters(formattedEvent)
```
BenFradet: isn't `reformatParameters` unnecessarily called twice when the event type is `user_updated`?

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/VeroAdapter.scala
```diff
@@ -0,0 +1,167 @@
...
+          context     = payload.context
+        )
+      }
+
+    } catch {
+      case e: JsonParseException => {
+        val exception = JU.stripInstanceEtc(e.toString).orNull
+        s"$VendorName event failed to parse into JSON: [$exception]".failureNel
+      }
+    }
```
BenFradet: I don't think we need to wrap the whole thing in a try/catch but just `parse`? or am I wrong (cc @chuwy)

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/VeroAdapter.scala
```diff
@@ -0,0 +1,167 @@
...
+   * @param json The event JSON which we need to
+   *        update values for
+   * @return the updated JSON with updated fields and values
+   */
+  private[registry] def reformatParameters(json: JValue): JValue = {
+
+    def toStringField(value: Long): JString = {
+      val dt: DateTime = new DateTime(value)
+      JString(JsonSchemaDateTimeFormat.print(dt))
+    }
```
BenFradet: can this fail? if so the result should ideally be a `Validation[E, JString]`

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/VeroAdapter.scala
```diff
@@ -0,0 +1,167 @@
...
+  /**
+   * Returns an updated Vero event JSON where
+   * the "action" field has been removed
+   * and "triggered_at" fields' values have been converted
+   *
+   * @param json The event JSON which we need to
+   *        update values for
+   * @return the updated JSON with updated fields and values
+   */
+  private[registry] def reformatParameters(json: JValue): JValue = {
```
BenFradet: no need for package private here

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/VeroAdapter.scala
misterpig: oops, the case matching should actually be:
```scala
case Some("user_updated") => parsed
```
`reformatParameters` is called once on all cases of `formattedEvent` to convert the field "triggered_at".

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/VeroAdapter.scala
misterpig: @chuwy made a similar comment I THINK in the Marketo adapter but I'm not sure what needs to be changed

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/VeroAdapter.scala
misterpig: this function was "borrowed" from the `cleanupJsonEventValues` method as is
the function is called only when it matches the case of
```scala
case ("triggered_at", JInt(x)")
```
so I'm not sure if any possible errors could arise if some integer values exist for that field to be converted
let me know what you guys think

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/VeroAdapter.scala
BenFradet:
```scala
val parsed = Try(parse(bodyJson)
val eventType = ...
```

instead of

```scala
try {
  val parsed = parse(bodyJson)
  val eventType = ...
} catch { ... }
```

chuwy: Hey @BenFradet, do we have a function in SCE codebase that parses string into `Validation[Something, JValue]`, just to show @misterpig how ideally it would look like.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/VeroAdapter.scala
```diff
@@ -0,0 +1,167 @@
...
+   * @return a Validation boxing either a NEL of RawEvents on
+   *         Success, or a NEL of Failure Strings
+   */
+  private def payloadBodyToEvent(json: String, payload: CollectorPayload): Validated[RawEvent] = {
+
+    val parsed = Try(parse(json))
+    parsed match {
+      case Success(parsed) => parsed
+      case Failure(e)      => return s"$VendorName event failed to parse into JSON: [${e.getMessage}]".failureNel
+    }
```
BenFradet: the result of the matchmaking is discarded, it should be:

```scala
val parsed = Try(parse(json)) match {
  case Success(parsed) => parsed
  case Failure(e) => s"$VendorName event failed to parse into JSON: [${e.getMessage}]".failureNel
}
```

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/VeroAdapter.scala
```diff
@@ -0,0 +1,167 @@
...
+    val eventType = (parsed.get \ "type").extractOpt[String].orElse(Some("user_updated"))
```
BenFradet: you should be able to do:

```scala
val eventType = parsed.toOption.flatMap(p => (p \ "type").extractOpt[String]).getOrElse("user_updated")
```

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/VeroAdapter.scala
```diff
@@ -0,0 +1,167 @@
...
+    val formattedEvent = eventType match {
+      case Some("user_updated") => parsed.get
+      case _ => {
+        cleanupJsonEventValues(
+          parsed.get,
+          ("type", eventType.get).some,
+          s"${eventType.get}_at"
+        )
+      }
+    }
```
BenFradet: you should be able to do:

```scala
val formattedEvent = parsed.map { p =>
  if (eventType == "user_updated") p
  else cleanupJsonEventValues(p, ("type", eventType).some, s"$eventType_at")
}
```

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/VeroAdapter.scala
```diff
@@ -0,0 +1,167 @@
...
+
+    val j1 = json.transformField {
+      case ("triggered_at", JInt(value)) => ("triggered_at", toStringField(value.toLong * 1000))
+    }
+    val j2 = j1.removeField {
+      case ("action", _) => true
+      case _             => false
+    }
+    j2
```
BenFradet: you can chain things:

```scala
json.transformField {
  case ("triggered_at", JInt(value)) => ("triggered_at", toStringField(value.toLong * 1000))
}.removeField {
  case ("action", _) => true
  case _ => false
}
```

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/VeroAdapter.scala
BenFradet: In general, it's good practice to avoid `.get`

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/VeroAdapter.scala
BenFradet: I think we're good

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/VeroAdapter.scala
misterpig: Heyo @BenFradet , I found that if I don't "return" early here on failure then running the test specs spits out an error instead of a failure.
E.g. passing in a bad json payload "hello" when tested shows:
```
! toRawEvents must return a success for a valid for all event types payload body being passed
[error]  JsonParseException: : Unrecognized token 'hello': was expecting ('true', 'false' or 'null')
[error]  at [Source: java.io.StringReader@42b5aa30; line: 1, column: 11]  (JsonParser.java:1524)
```
versus
```
x toRawEvents must return a success for a valid for all event types payload body being passed
[error]  'Failure(NonEmptyList(Twilio event failed to parse into JSON: [Unrecognized token 'hello': was expecting ('true', 'false' or 'null')
[error]   at [Source: java.io.StringReader@38849b9f; line: 1, column: 11]]))' is not Success with value'NonEmptyList(...
```

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/VeroAdapter.scala
misterpig: Similarly here, 'formattedEvent' will return wrapped in Try() but needs to be a JValue to be passed into the function 'reformatParameters' and then into 'toUnstructEventParams'.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/VeroAdapter.scala
BenFradet: you don't need to return early you just need to carry on with your failure, this is why there is a `parsed.map` in my comment below because `parsed` will have type `ValidationNel[String, JValue]`.

by the way, the comment above should say `case Success(parsed) => parsed.successNel`, that's my bad.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/VeroAdapter.scala
misterpig: I'm running into some trouble here. The val 'eventType' is a string which will fail when passed into lookupSchema as it requires a Option[] object. I could get around this by using
`lookupSchema(Some(eventType)...`

BenFradet: yup, you can also do `eventType.some` too

Here you don't it to be an option and you can have it be a simple string so it's better to go towards the simpler type

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/VeroAdapter.scala
BenFradet: you can do a for comprehension:

```scala
val parsed = Try(parse(json)) match {
  case Success(p) => p.successNel
  case Failure(e) => s"$VendorName event failed to parse into JSON: [${e.getMessage}]".failureNel
}
val eventType = parsed.toOption.flatMap(p => (p \ "type").extractOpt[String]).getOrElse("user_updated")
for {
  parsedEvent <- parsed
  formattedEvent =
    if (eventType == "user_updated") parsedEvent
    else cleanupJsonEventValues(parsedEvent, ("type", eventType).some, s"$eventType_at")
  reformattedEvent = reformatParameters(formattedEvent)
  params = toUnstructEventParams(TrackerVersion, toMap(payload.queryString), schema, reformattedEvent, "srv")
  schema <- lookupSchema(eventType.some, VendorName, EventSchemaMap).toValidationNel
  rawEvent = RawEvent(api = payload.api, parameters = params, contentType = payload.contentType, source = payload.source, context = payload.context)
} yield rawEvent
```

BenFradet:
Integrated into #3799, thanks so much, closing :+1:

PR #3716, opened 2018-04-10
Scala Common Enrich: bump mandrill event versions to 1-0-1

BenFradet:
blocked by snowplow/iglu-central#763


alexanderdean:
I don't think this is blocked - we can emit the newer schemas, they will pass validation, we simply will be loading the older versions (1-0-0s) into Redshift.

PR #3734, opened 2018-04-18
Scala Common Enrich: propagate the currency code to all the contexts …

### LINKED ISSUE #3733: Scala Common Enrich: propagate the currency code to all the contexts which need it in the GA adapter
Instead of treating `cu` as a composite context and get a mismatch between the number of currency code composite contexts (1) and the actual number of composite contexts which need it, we should attach `currencyCode` to all the composite contexts which need it after the fact.

BenFradet:
…which need it in the GA adapter (closes #3733)


BenFradet:
@chuwy, the original problem was that only one "currency code composite context" was built but if there were multiple composite contexts needing the currency code (e.g. multiple product impression) only one would receive the currency code.

What this PR does is attach the currency code to all the composite contexts which need it after the fact so that every context which needs it gets it.

Additionally, there is a small improvement into how the raw params are transformed before being translated to our schemas so that the data is rectangular which facilitates transposition.

BenFradet:
In #3799 closing

PR #3747, opened 2018-04-30
Release R105

BenFradet:
- [x] wiki date
- [x] cl date
- [x] bp date


3-enrich/stream-enrich/kinesis/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/sinks/KinesisSink.scala
```diff
@@ -41,12 +41,7 @@ import scalatracker.Tracker
 
 /** KinesisSink companion object with factory method */
 object KinesisSink {
-  def createAndInitialize(
-    kinesisConfig: Kinesis,
-    bufferConfig: BufferConfig,
-    streamName: String,
-    tracker: Option[Tracker]
-  ): \/[String, KinesisSink] = for {
+  def validate(kinesisConfig: Kinesis, streamName: String): \/[String, Boolean] = for {
```
chuwy: For me it looks like `\/[String, Unit]` rather than `Boolean` - it'll never return `false`.

3-enrich/stream-enrich/kinesis/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/sinks/KinesisSink.scala
```diff
@@ -55,12 +50,12 @@ object KinesisSink {
       .withCredentials(provider)
       .withEndpointConfiguration(endpointConfiguration)
       .build()
-    _ <- streamExists(client, streamName).leftMap(_.getMessage)
+    exists <- streamExists(client, streamName).leftMap(_.getMessage)
       .flatMap { b =>
```
chuwy: This `flatMap` can be changed to `streamExists(client, streamName).ensure("Kinesis stream $streamName doesn't exist")(_ == true)`

3-enrich/stream-enrich/kinesis/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/sources/KinesisSource.scala
```diff
@@ -56,37 +58,43 @@ object KinesisSource {
...
-    threadLocalGoodSink = new ThreadLocal[Sink] {
-      override def initialValue = goodSink
-    }
-    badSink <-
-      KinesisSink.createAndInitialize(kinesisConfig, config.buffer, config.out.bad, tracker)
-        .validation
-    threadLocalBadSink = new ThreadLocal[Sink] {
-      override def initialValue = badSink
-    }
+    _ <- KinesisSink.validate(kinesisConfig, config.out.enriched).validation
```
chuwy: Don't we want to aggregate these? If `validate` fails on `config.out.enriched` we'll never know that `config.out.bad` is also invalid.

BenFradet:
@chuwy thanks, updated!

3-enrich/stream-enrich/nsq/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/sources/NsqSource.scala
```diff
@@ -90,14 +87,14 @@ class NsqSource private (
 
     val errorCallback = new NSQErrorCallback {
       override def error(e: NSQException): Unit =
-        log.error(s"Exception while consuming topic $topicName", e)
+        log.error(s"Exception while consuming topic $${config.in.raw}", e)
```
chuwy: Is this `$` redundant?

knservis: It is there to escape $ which would cause an error. The s" is needed to silence a warning.

BenFradet:
@knservis the summary in https://discourse.snowplowanalytics.com/t/important-alert-r101-bug-may-result-in-duplicated-data-in-the-real-time-pipeline/1987 details the issue.

PR #3832, opened 2018-07-18
#3831 Fix that adds \n at the end of every event

### LINKED ISSUE #3831: Redshift copy failure due to missing break-lines
Hi guys! I currently have a stream-enricher which is generating and publishing events (TSV's) to a Kinesis topic. I have a Kinesis Firehose connected to a Kinesis topic, and the problem I'm facing is that the generated file is a single line document (all tsv's records are appended in the same line). For this reason when Firehose tries to do a COPY command to Redshift, it fails.

I have a fix ready for this, and I can submit a PR if it makes sense to you guys. But I'd like to make sure I'm not doing anything wrong and the approach I'm following is right.

Please, let me know if you have any questions or concerns. Thanks a lot in advance for you guidance!

anistal:
It closes #3831 adding a '\n' to the end of the final event (that it is a TSV)


alexanderdean:
Closing - this is not a fix.

PR #3836, opened 2018-07-20
R110 Vallei Dei Templei

BenFradet:
- [ ] Add snowplow to https://github.com/spotify/scio/wiki/Powered-By


BenFradet:
superseded by #3890

3-enrich/beam-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.beam/Enrich.scala
```diff
@@ -0,0 +1,205 @@
...
+        .flatten.withName("pii-successes")
+        .partition(_._2 >= MaxRecordSize)
+      Some((tooBigPiis, properlySizedPiis))
+    } else {
+      None
+    }
+
+    (piis |@| config.pii) { (piis, pii) =>
+      val properlySizedPiis = piis._2
+      properlySizedPiis.withName("undersized-pii-successes")
```
knservis: Undersized sounds like there is something wrong with them (unless I misunderstood the code and there is something wrong)

3-enrich/beam-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.beam/utils.scala
```diff
@@ -0,0 +1,167 @@
...
+
+  /**
+   * Truncate an oversized formatted enriched event into a [[BadRow]].
+   * @param enrichedEvent TSV-formatted oversized enriched event
+   * @param bytesSize size in bytes of the formatted enriched event
+   * @param maxBytesSize maximum size in bytes a record can take
+   * @return a [[BadRow]] containing a the truncated enriched event (10 times less than the max size)
+   */
+  def resizeEnrichedEvent(enrichedEvent: String, bytesSize: Int, maxBytesSize: Int): BadRow =
+    BadRow(enrichedEvent.take(maxBytesSize / (4 * 10)), NonEmptyList(
```
knservis: I don't understand this number (/(4*10)) in this context (I am thinking of 39 max bytes for instance). Could we make it a constant with a descriptive name? Also, would it not be better to guarantee some fields make it to the truncated BadRow (for instance eventId)?

3-enrich/beam-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.beam/utils.scala
```diff
@@ -0,0 +1,167 @@
...
+      s"Size of enriched event ($bytesSize) is greater than allowed maximum ($maxBytesSize)"))
+
+  /**
+   * Resize a [[BadRow]] if it exceeds the maximum allowed size.
+   * @param badRow the original [[BadRow]] which can be oversized
+   * @param maxBytesSize maximum size in bytes a record can take
+   * @return a [[BadRow]] where the line is 10 times less than the max size
+   */
+  def resizeBadRow(badRow: BadRow, maxBytesSize: Int): BadRow = {
```
knservis: It looks like this should be handled by the def above `resizeEnrichedEvent`

3-enrich/beam-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.beam/utils.scala
```diff
@@ -0,0 +1,167 @@
...
+      badRow.copy(
+        line = badRow.line.take(maxBytesSize / (4 * 10)),
+        errors = msg.toProcessingMessage <:: badRow.errors
+      )
+    } else badRow
+  }
+
+  /** Get the size of a string in bytes. */
+  def getStringSize(string: String): Int =
+    ByteBuffer.wrap(string.getBytes(UTF_8)).capacity
```
knservis: Is that different than `string.getBytes(UTF_8).size` ?

3-enrich/beam-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.beam/utils.scala
```diff
@@ -0,0 +1,167 @@
...
+   * @return either the path of the created sym link or the error
+   */
+  def createSymLink(file: File, symLink: String): Either[String, Path] = {
+    val symLinkPath = Paths.get(symLink)
+    if (Files.notExists(symLinkPath)) {
+      Try(Files.createSymbolicLink(symLinkPath, file.toPath)) match {
+        case scala.util.Success(p) => Right(p)
+        case scala.util.Failure(t) => Left(s"Symlink can't be created: ${t.getMessage}")
+      }
+    } else Left(s"Symlink $symLinkPath already exists")
```
knservis: This entry may exist but that does not mean it is a symlink. Also as per documentation notExists: "method is not the complement of the exists method. Where it is not possible to determine if a file exists or not then both methods return false."

2-collectors/clojure-collector/java-servlet/project.clj
```diff
@@ -13,7 +13,7 @@
 ;;;; Copyright: Copyright (c) 2012-2013 Snowplow Analytics Ltd
```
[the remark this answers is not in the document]
BenFradet: I'll log a separate ticket

3-enrich/beam-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.beam/Enrich.scala
```diff
@@ -0,0 +1,205 @@
...
+    ScioMetrics.distribution(MetricsNamespace, "enriched_event_size_bytes")
+  val timeToEnrichDistribution =
+    ScioMetrics.distribution(MetricsNamespace, "time_to_enrich_ms")
+
+  def main(cmdlineArgs: Array[String]): Unit = {
+    val (sc, args) = ContextAndArgs(cmdlineArgs)
+    val parsedConfig = for {
+      config <- EnrichConfig(args)
+      _ = sc.setJobName(config.jobName)
+      _ <- checkTopicExists(sc, config.output)
```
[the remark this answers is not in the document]
BenFradet: I've actually renamed it to enriched to fit https://github.com/snowplow/snowplow/blob/master/3-enrich/stream-enrich/examples/config.hocon.sample

3-enrich/beam-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.beam/utils.scala
BenFradet: I thought so but apparently not

3-enrich/beam-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.beam/utils.scala
BenFradet: good point, I added a comment to the constant saying: "We want to take one-tenth of the payload characters and one character can take up to 4 bytes"

3-enrich/beam-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.beam/utils.scala
BenFradet: there are two different methods because for enriched events that are too big we already know their size whereas for bad rows we don't

3-enrich/beam-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.beam/utils.scala
BenFradet: woops, wrong comment

3-enrich/beam-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.beam/utils.scala
BenFradet: good point, I added a comment to the constant saying: "We want to take one-tenth of the payload characters and one character can take up to 4 bytes"

Issue #3793, opened 2018-06-07
Scala Common Enrich: Externalize ua-parser rule file


userkci:
The ua-parser enrichment uses a rule file which is embedded in the uber-jar.  However, this is dependent on which the rule file happens to be available at build time.  Currently, this is brought in via the ua-parser client jar which dates from 2013.

It would be useful for the rule file to be a configurable setting for the enrichment, similar to the geoip enhancement.  This way, the rules can be updated independent of ua-parser or Snowplow release.

As a workaround, one can replace the rules file in the jar (ua_parser/regexes.yaml) and use the modified jar in Snowplow ETL configuration.

BenFradet:
This is actually what we are moving towards with the referer parser enrichment and it does also make sense for ua-parser :+1:

userkci:
It looks to me that you can initialize the parser with an InputStream, so I don't think a PR is required for ua-parser.

We have a need for this change and I'll have some free time in the next week, so hopefully can provide a PR for this change shortly.

Issue #3789, opened 2018-06-05
Scala Common Enrich: make the list of files to cache available from the EnrichmentRegistry


BenFradet:
As the numbers of local files to cache proliferate (e.g. #3750) and the number of enrichment platforms increases (e.g. #3736).

I think it should be the responsibility of the `EnrichmentRegistry` to tell us which files need caching.

PR #3853, opened 2018-08-01
R109 Lambaesis

BenFradet:
Renamed from Mileum, supersedes #3824


BenFradet:
>Are we doing a new release of Spark Enrich, or not?

Not but a couple of tickets are applicable to both:

- #3789
- #3793

And it made a lot of sense to add them now instead of, for example, in 4 releases where batch will get all the features we've accumulated, that's why there is a sce bump.

To sum up, it will greatly ease the next batch release.

alexanderdean:
Orthogonal: in this file:

https://github.com/snowplow/snowplow/blob/release/r109-lambaesis/3-enrich/config/enrichments/ua_parser_config.json#L10

Can we rename:

```
regexes.yaml
```

to:

```
useragents-latest.yaml
```

regexes is too vague.

BenFradet:
@darrenhaken unfortunately no, it's named after [the Algerian archaeological site](https://en.wikipedia.org/wiki/Lambaesis) :smile:

BenFradet:
@alexanderdean 

>Can we rename ...

We've renamed it to `regexes-latest.yaml` to be consistent with the monthly thing we want to have, i.e. a latest and a yyyy-MM. However, the original dataset is actually named [`regexes.yaml`](https://github.com/ua-parser/uap-core/blob/master/regexes.yaml) so I don't know if we'd want to deviate from that too much.

alexanderdean:
> original dataset is actually named `regexes.yaml`

Fair play, let's stick with that and keep the origins clear...

PR #3881, opened 2018-08-24
Scala Common Enrich: move away from user-agent-utils (closes #3591)

### LINKED ISSUE #3591: Scala Common Enrich: remove user-agent-utils enrichment
The page on bitwalker says this project is being end of lifed
http://www.bitwalker.eu/software/user-agent-utils
Will need to move to an alternate.

szareiangm:
This is the first step discussed in the ticket. The next steps (refactoring events table and stop populating the 10 fields that are generated by this enrichment) would not fit in this ticket. So I think this PR will close the issue, too. 

Making sure that this enrichment and its fields are gone is important for us since they count for ~%8 of all main table fields and the volume of the data is significant when considering larger scale.

CC: @nafid


snowplowcla:
@szareiangm  has signed the Software Grant and Corporate Contributor License Agreement

BenFradet:
integrated into #3965 , thanks a lot! :+1:

Issue #3822, opened 2018-07-02
Redshift: widen geo_region to 3 characters


miike:
This is related to shredding/data modelling but putting here for visibility.

Recently Maxmind has started including more ISO-3166-2 region codes in their GeoIP lookup databases as opposed to older FIPS 10-4 region codes of only two characters. Many of these ISO-8166-2 region codes are three characters in length rather than two.

At the moment the pipeline truncates the region (the [ISO code](https://github.com/snowplow/scala-maxmind-iplookups/blob/61ef850b67aff0b550c0f8805e37eaa1c5b1e062/src/main/scala/com.snowplowanalytics.maxmind.iplookups/model.scala#L46) for the most specific subdivision) in two places which currently results in a discrepancy in enriched data (with full ISO-3166-2 codes) and the shredded data (where the ISO code has been truncated).

Currently this occurs in two places:
- The logic that enforces the Postgres constraints in [rdb-shredder](https://github.com/snowplow/snowplow-rdb-loader/blob/2343b1d512b6997b8b9cc4b81b238b996688e756/shredder/src/main/scala/com.snowplowanalytics.snowplow.storage/spark/ShredJob.scala#L199) and
- The atomic.events model for both Redshift and Postgres itself which has historically always been a `char(2)` (of which `TRUNCATECOLUMNS` will prevent three character codes being inserted).

In some countries this is less of an issue but for others the truncation can introduce ambiguity, e.g., in GB
```
SWK => Southwark
SWD => Swindon
SWA => Swansea
```

The Postgres enforcement logic in rdb-shredder and the atomic.events model should both be updated to accept three character region codes.

BenFradet:
Excellent find, I logged snowplow/snowplow-rdb-loader#110 :+1:

PR #3903, opened 2018-09-27
Bump the length of geo_region from 2 to 3 characters

miike:
This addresses https://github.com/snowplow/snowplow/issues/3822 for both Redshift and Postgres including the relevant migration scripts.

I'll make a new PR for https://github.com/snowplow/snowplow-rdb-loader/issues/110


snowplowcla:
@miike  has signed the Software Grant and Corporate Contributor License Agreement

4-storage/redshift-storage/sql/atomic-def.sql
```diff
@@ -9,10 +9,10 @@
 -- "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 -- See the Apache License Version 2.0 for the specific language governing permissions and limitations there under.
 --
--- Version:     0.10.0
+-- Version:     0.11.0
```
BenFradet: should this be 0.10.1, to mirror postgres?

miike: I'm not too sure on the semantics of this. For Redshift we need to create an entirely new table whereas for Postgres we can just run an `ALTER COLUMN`.

chuwy: I think Redshift never had anything in PATCH except 0, even when only ALTER TABLE was enough. I still think that these versions are preferable. At least if we're going to change something - let's change Postgres.

BenFradet: @chuwy so should we change postgres?

4-storage/postgres-storage/sql/migrate_0.7.0_to_0.7.1.sql
```diff
@@ -0,0 +1,26 @@
...
+-- URL:         -
+--
+-- Authors:     Mike Robins
+-- Copyright:   Copyright (c) 2018 Snowplow Analytics Ltd
+-- License:     Apache License Version 2.0
+
+BEGIN TRANSACTION;
+
+ALTER TABLE "atomic"."events"
+    ALTER COLUMN geo_region TYPE varchar(3);
```
chuwy: Just noticed. Why we're adding `varchar` whereas in DDL it is `char`?

4-storage/postgres-storage/sql/migrate_0.7.0_to_0.7.1.sql
```diff
@@ -0,0 +1,26 @@
...
+-- See the Apache License Version 2.0 for the specific language governing permissions and limitations there under.
+--
+-- Version:     Ports version 0.7.0 to version 0.7.1
+-- URL:         -
+--
+-- Authors:     Mike Robins
+-- Copyright:   Copyright (c) 2018 Snowplow Analytics Ltd
+-- License:     Apache License Version 2.0
+
+BEGIN TRANSACTION;
```
chuwy: I actually don't think transaction works for `ALTER COLUMN`

4-storage/redshift-storage/sql/atomic-def.sql
chuwy: I'm in doubt as I also don't know what are semantics of these versions (it isn't SchemaVer for sure). But I think yes - let's change postgres.

4-storage/postgres-storage/sql/migrate_0.7.0_to_0.7.1.sql
miike: Good catch, I've changed this.

4-storage/postgres-storage/sql/migrate_0.7.0_to_0.7.1.sql
miike: I tried this in 10.5 and it seemed to work. Should I try this on other versions?

chuwy: I just mean that it probably won't be a transaction, e.g. if `COMMENT` will fail (let's imagine) `ALTER COLUMN` won't be rolled back.

miike: I could be incorrect but I though that [Postgres DDLs](https://wiki.postgresql.org/wiki/Transactional_DDL_in_PostgreSQL%3a_A_Competitive_Analysis) are transactional so this would be rolled back.

chuwy: Okay, I'm actually happy to be wrong here.

Issue #3904, opened 2018-10-02
Scala Stream Collector: add support for an Access-Control-Max-Age header


miike:
Allow the OPTIONS request to return a configurable `Access-Control-Max-Age` header to cache the returned response from the collector.

This results in:
- faster network requests (no requirement to wait for OPTIONS)
- fewer OPTIONS requests depending on cache setting

PR #3905, opened 2018-10-02
Add a configurable Access-Control-Max-Age header

miike:
This adds an additional header to allow caching of the OPTIONS response from the collector see #3904 

@BenFradet Where do you think I should put this property in the configuration hocon? I'm not sure that putting it at the root of the collector config is the best option.

### DIFF
--- a/2-collectors/scala-stream-collector/config.hocon.sample
+++ b/2-collectors/scala-stream-collector/config.hocon.sample
@@ -1,0 +1,165 @@
+# Copyright (c) 2013-2016 Snowplow Analytics Ltd. All rights reserved.
+#
+# This program is licensed to you under the Apache License Version 2.0, and
+# you may not use this file except in compliance with the Apache License
+# Version 2.0.  You may obtain a copy of the Apache License Version 2.0 at
+# http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the Apache License Version 2.0 is distributed on an "AS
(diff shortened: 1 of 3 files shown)


miike:
One of the tests is failing at the moment. I'll fix this once we've confirmed where we want to put the configuration option.

snowplowcla:
@miike  has signed the Software Grant and Corporate Contributor License Agreement

2-collectors/scala-stream-collector/config.hocon.sample
```diff
@@ -0,0 +1,165 @@
...
+  remote-address-header = on
+
+  uri-parsing-mode = relaxed
+  raw-request-uri-header = on
+
+  # Define the maximum request length (the default is 2048)
+  parsing {
+    max-uri-length = 32768
+  }
+}
```
BenFradet: please modify the existing config: https://github.com/snowplow/snowplow/blob/master/2-collectors/scala-stream-collector/examples/config.hocon.sample

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala
```diff
@@ -113,6 +113,7 @@ package model {
   final case class CollectorConfig(
     interface: String,
     port: Int,
+    accessControlMaxAge: Int,
```
BenFradet: should this be optional?

BenFradet:
hey @miike , could you update the PR?

BenFradet:

Issue #3421, opened 2017-09-07
Scala Stream Collector: expose Prometheus metrics


christoph-buente:
In order to monitor the parts of the real-time stack, it would be very beneficial to have a way to monitor each step of the pipeline. For example, how many requests got into the collector, how much got out to kinesis, how many instances are running in each stage? I know we can send some metrics into the collector itself, but it feels wrong to send data to an endpoint, that i actually want to monitor itself.

I would really like to see support for prometheus for collector, enricher and storage sinks. All it requires is a http server listening to a given port and expose some metrics in pre-defined text format. 

What do you think?

https://github.com/prometheus/prometheus

alexanderdean:
Yes good idea, we have had customers asking for this... Any suggestions on the metrics?

christoph-buente:
For the collector, i think the usual suspects for any http endpoint:
* http_requests_total{path, code, method}
* http_request_duration_seconds_count
* http_request_duration_seconds_sum
* http_request_duration_seconds_bucket
* http_request_size_bytes
* http_request_size_bytes_count
* http_request_size_bytes_sum

Also for all parts reading from and writing to kinesis/kafka
* kinesis_events_write_total{type,format,retry,failure/success}
* kinesis_events_read_total{type,format,retry,failure/success}

For all parts in common:
* build_info{scalaversion,javaversion,version,gitsha}

Add anything that the parts send as data metrics to the collector right now.

PR #3929, opened 2018-11-16
Scala stream collector: Expose prometheus metrics (#3421)

melgenek:
This PR exposes Scala collector prometheus metrics as discussed in the ticket https://github.com/snowplow/snowplow/issues/3421

The example response from `/metrics` endpoint that can be fed into prometheus:
```
# HELP http_request_duration_seconds Latency per endpoint
# TYPE http_request_duration_seconds histogram
http_request_duration_seconds_bucket{endpoint="/i",method="GET",code="200",le="0.005",} 155.0
http_request_duration_seconds_bucket{endpoint="/i",method="GET",code="200",le="0.01",} 156.0
http_request_duration_seconds_bucket{endpoint="/i",method="GET",code="200",le="0.025",} 156.0
http_request_duration_seconds_bucket{endpoint="/i",method="GET",code="200",le="0.05",} 157.0
http_request_duration_seconds_bucket{endpoint="/i",method="GET",code="200",le="0.075",} 157.0
http_request_duration_seconds_bucket{endpoint="/i",method="GET",code="200",le="0.1",} 157.0
http_request_duration_seconds_bucket{endpoint="/i",method="GET",code="200",le="0.25",} 158.0
http_request_duration_seconds_bucket{endpoint="/i",method="GET",code="200",le="0.5",} 158.0
http_request_duration_seconds_bucket{endpoint="/i",method="GET",code="200",le="0.75",} 158.0
http_request_duration_seconds_bucket{endpoint="/i",method="GET",code="200",le="1.0",} 158.0
http_request_duration_seconds_bucket{endpoint="/i",method="GET",code="200",le="2.5",} 158.0
http_request_duration_seconds_bucket{endpoint="/i",method="GET",code="200",le="5.0",} 158.0
http_request_duration_seconds_bucket{endpoint="/i",method="GET",code="200",le="7.5",} 158.0
http_request_duration_seconds_bucket{endpoint="/i",method="GET",code="200",le="10.0",} 158.0
http_request_duration_seconds_bucket{endpoint="/i",method="GET",code="200",le="+Inf",} 158.0
http_request_duration_seconds_count{endpoint="/i",method="GET",code="200",} 158.0
http_request_duration_seconds_sum{endpoint="/i",method="GET",code="200",} 0.45800000000000024
# HELP service_version Java, scala versions and runnable name
# TYPE service_version gauge
service_version{java_version="1.8.0_191",scala_version="2.11.11",runnable="snowplow-stream-collector-stdout-0.14.1.jar",} 1.0
# HELP http_requests_total Total count of requests to http endpoint
# TYPE http_requests_total counter
http_requests_total{endpoint="/i",method="GET",code="200",} 158.0
```


christoph-buente:
@alexanderdean and @BenFradet 
i submitted another corporate CLA for our new engineers including @melgenek

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/metrics/MetricsDirectives.scala
```diff
@@ -0,0 +1,23 @@
+package com.snowplowanalytics.snowplow.collectors.scalastream.metrics
```
BenFradet: can you add the apache header to the different files you've added? :+1:

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/metrics/MetricsService.scala
```diff
@@ -0,0 +1,68 @@
...
+  final val NanosecondsInSecond: Double = Math.pow(10, 9)
+
+  object Metrics {
+    final val HttpRequestDuration = "http_request_duration_seconds"
+    final val HttpRequestDurationHelp = "Latency per endpoint"
+
+    final val HttpRequestCount = "http_requests_total"
+    final val HttpRequestCountHelp = "Total count of requests to http endpoint"
+
+    final val Labels = Seq("endpoint", "method", "code")
```
BenFradet: no need for `final` in objects :+1:

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/metrics/MetricsService.scala
```diff
@@ -0,0 +1,68 @@
...
+    .labelNames("java_version", "scala_version", "runnable")
+    .register(registry)
+
+  applicationVersionGauge.labels(StaticMetrics.javaVersion, StaticMetrics.scalaVersion, StaticMetrics.runnableName).set(1)
+
+  override def observeRequest(method: HttpMethod, uri: Uri, status: StatusCode, duration: Duration): Unit = {
+    val path = uri.path.toString
+    val methodValue = method.value
+    val code = status.intValue().toString
+    requestDurationHistogram.labels(path, methodValue, code).observe(duration.toNanos / NanosecondsInSecond)
```
BenFradet: why not just `toSeconds` and getting rid of the division?

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/metrics/MetricsDirectives.scala
```diff
@@ -0,0 +1,23 @@
...
+
+import java.time.{Duration, LocalDateTime}
+
+import akka.http.scaladsl.server.Directive0
+import akka.http.scaladsl.server.Directives._
+import akka.http.scaladsl.server.directives.BasicDirectives.extractRequestContext
+
+trait MetricsDirectives {
```
BenFradet: also if you could add Scaladoc to the different public members that'd be much appreciated :+1:

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/metrics/MetricsService.scala
```diff
@@ -0,0 +1,68 @@
...
+  private val requestDurationHistogram: Histogram =
+    metricsConfig.durationBuckets
+      .map(buckets => requestDurationHistogramBuilder.buckets(buckets: _*).register(registry))
+      .getOrElse(requestDurationHistogramBuilder.register(registry))
+
+  private val applicationVersionGauge = Gauge.build("service_version", "Java, scala versions and runnable name")
+    .labelNames("java_version", "scala_version", "runnable")
+    .register(registry)
+
+  applicationVersionGauge.labels(StaticMetrics.javaVersion, StaticMetrics.scalaVersion, StaticMetrics.runnableName).set(1)
```
BenFradet: Isn't there a better metric type than gauge for those metrics?

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/metrics/StaticMetrics.scala
```diff
@@ -0,0 +1,21 @@
...
+
+object StaticMetrics {
+
+  def javaVersion: String = classOf[Runtime].getPackage.getImplementationVersion
+
+  def scalaVersion: String = Try {
+    val props = new java.util.Properties
+    props.load(getClass.getResourceAsStream("/library.properties"))
+    props.getProperty("version.number")
+  }.getOrElse("")
```
BenFradet: no need for all this reflection, you can leverage a sbt plugin we're already using to expose this build info: https://github.com/sbt/sbt-buildinfo

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/metrics/StaticMetrics.scala
```diff
@@ -0,0 +1,21 @@
...
+
+  def runnableName: String = Try {
+    new File(getClass.getProtectionDomain.getCodeSource.getLocation.getPath).getName
+  }.getOrElse("")
```
BenFradet: this can just be https://github.com/snowplow/snowplow/blob/08d2f51b44148169427e5f423f7ea4dd489af9e5/2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala#L73-L74 no need for the jar name

2-collectors/scala-stream-collector/examples/config.hocon.sample
```diff
@@ -180,6 +180,16 @@ collector {
       timeLimit = {{bufferTimeThreshold}}
     }
   }
+
+  # Configuration of prometheus http metrics
+  metrics {
```
BenFradet: should this section be named `prometheusMetrics`?

2-collectors/scala-stream-collector/examples/config.hocon.sample
```diff
@@ -180,6 +180,16 @@ collector {
...
+    # If metrics are enabled then all requests will be logged as prometheus metics
```
BenFradet: metics

2-collectors/scala-stream-collector/examples/config.hocon.sample
```diff
@@ -180,6 +180,16 @@ collector {
...
+    # and '/metrics' endpoint will return the report about the requests
+    enabled = false
+    # Custom buckets for http_request_duration_seconds_bucket duration metric
+    # durationBuckets = [0.1, 3, 10]
```
BenFradet: this needs to be uncommented otherwise configuration parsing won't work :+1:

2-collectors/scala-stream-collector/project/Dependencies.scala
```diff
@@ -55,6 +56,8 @@ object Dependencies {
     val slf4j                = "org.slf4j"             %  "slf4j-simple"           % V.slf4j
     val log4jOverSlf4j       = "org.slf4j"             %  "log4j-over-slf4j"       % V.slf4j
     val config               = "com.typesafe"          %  "config"                 % V.config
+    val prometheus           = "io.prometheus"         %  "simpleclient"           % V.prometheus
+    val prometheus_common    = "io.prometheus"         %  "simpleclient_common"    % V.prometheus
```
BenFradet: let's use camel case :+1:

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/metrics/MetricsService.scala
melgenek: There is no public `toSeconds` and nanos give the best precision after division

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/metrics/MetricsService.scala
melgenek: Actually constant values are better to be declared using `final` modifier. 
https://www.scala-lang.org/files/archive/spec/2.12/05-classes-and-objects.html
`Note, however, that constant value definitions do require an explicit final modifier, even if they are defined in a final class or object.`

The final variables are set during compile time, while non-final members of object are just method calls.
Consider such example
```
object Sample {
  def main(args: Array[String]): Unit = {
    println(Constants.str1)
    println(Constants.str2)
  }
}

object Constants {
  val str1 = "String1"
  final val str2 = "String2"
}
```
Then compile `scalac -print Sample.scala`

And you can see the difference
```
[[syntax trees at end of                   cleanup]] // Sample.scala
package <empty> {
  object Sample extends Object {
    def main(args: Array[String]): Unit = {
      scala.this.Predef.println(Constants.str1());
      scala.this.Predef.println("String2")
    };
    def <init>(): Sample.type = {
      Sample.super.<init>();
      ()
    }
  };
  object Constants extends Object {
    private[this] val str1: String = _;
    <stable> <accessor> def str1(): String = Constants.this.str1;
    final <stable> <accessor> def str2(): String("String2") = "String2";
    def <init>(): Constants.type = {
      Constants.super.<init>();
      Constants.this.str1 = "String1";
      ()
    }
  }
}
```

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/metrics/StaticMetrics.scala
melgenek: Thank you for the hint. I have left only the version, as it is only one `/metrics` per service and the name is not necessary

2-collectors/scala-stream-collector/examples/config.hocon.sample
melgenek: This is commented on purpose. 
By default built-in buckets are used. If they does not meet the needs of application then they can be configured via hocon.
The parameter `durationBuckets` is specified as optional and is allowed to be omitted

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/metrics/MetricsDirectives.scala
melgenek: added some info

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/metrics/MetricsService.scala
BenFradet: The problem is that I don't think neither nanos nor seconds are useful as a request duration unit of measurement. In my mind milliseconds would definitely be better suited, wdyt?

christoph-buente: Regarding to the best practices guidelines from prometheus, they encourage to use base metrics like seconds as a name for the metric. The value for milliseconds would then be a fraction of it. See here: https://prometheus.io/docs/practices/naming/#base-units

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/metrics/MetricsService.scala
christoph-buente: We followed advice from Brian Brazil, core maintainer of prometheus regarding the software version. Have a look here: https://www.robustperception.io/exposing-the-software-version-to-prometheus

christoph-buente:
@BenFradet Were we able to address all your concerns?

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/metrics/MetricsRoute.scala
```diff
@@ -0,0 +1,42 @@
...
+        entity = HttpEntity.Strict(MetricsRoute.CONTENT_TYPE_004, ByteString(metricsService.report()))
+      ))
+    }
+
+}
+
+
+object MetricsRoute {
+
+  val CONTENT_TYPE_004: ContentType = ContentType.parse(TextFormat.CONTENT_TYPE_004).right.get
```
BenFradet: can we avoid this get?

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/metrics/StaticMetrics.scala
BenFradet: No need for the reflection for both the java and scala versions

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala
```diff
@@ -14,11 +14,10 @@
  */
 package com.snowplowanalytics.snowplow.collectors.scalastream
 
-import scala.concurrent.duration.FiniteDuration
-
 import akka.http.scaladsl.model.headers.HttpCookie
+import com.snowplowanalytics.snowplow.collectors.scalastream.sinks.Sink
 
-import sinks.Sink
+import scala.concurrent.duration.FiniteDuration
```
BenFradet: :art: can you leave the imports as they were?

2-collectors/scala-stream-collector/core/src/test/scala/com.snowplowanalytics.snowplow.collectors.scalastream/metrics/MetricsServiceSpec.scala
```diff
@@ -0,0 +1,61 @@
...
+
+  "Prometheus metrics service" should {
+    "return report about the observed requests" in {
+      val metricsService = new PrometheusMetricsService(PrometheusMetricsConfig(enabled = true, None))
+
+      metricsService.observeRequest(HttpMethods.POST, Uri("/endpoint"), StatusCodes.BadGateway, Duration.ofSeconds(3))
+
+      val report = metricsService.report()
+
+      report contains
```
BenFradet: I think this should be `should include`

2-collectors/scala-stream-collector/core/src/test/scala/com.snowplowanalytics.snowplow.collectors.scalastream/metrics/MetricsServiceSpec.scala
```diff
@@ -0,0 +1,61 @@
...
+        """# HELP http_requests_total Total count of requests to http endpoint
+          |# TYPE http_requests_total counter
+          |http_requests_total{endpoint="/endpoint",method="POST",code="502",} 1.0""".stripMargin
+
+      report contains
```
BenFradet: same

christoph-buente:
Thank you @melgenek and @BenFradet for moving forward 🥇

BenFradet:

PR #3933, opened 2018-11-27
Add persistent EMR Cluster for EER runs

jbeemster:
Hey @BenFradet so this branch can now successfully push steps to a persistent cluster.  The logic is as follows:

1. If the `--use-persistent-jobflow` argument is passed to `run` then it will attempt to discover a jobflow with the name specified in the config and return the jobflow_id to attach to
  - If no jobflow exists with this name that is in a RUNNING or WAITING state then it simply creates a new cluster
  - If a jobflow is found but it is in RUNNING state it fails as it is currently processing
  - If more than one jobflow is found in RUNNING or WAITING state it fails as the jobflow must be unique
  - If a unique jobflow is found in WAITING then the id is returned and we attach to it
2. If a jobflow is found then all `application`, `configuration` and `bootstrap` actions are ignored as they cannot be applied to a running cluster
3. If a persistent cluster should be used then all steps have their failure action changed to `cancel_and_wait`

For parsing the job steps I am filtering based on the run_tstamp so we only look at the job steps from this run rather than all job steps.  From my experimentation the list_steps function in Elasticity returns the most recent 50 steps so we should not need to do any pagination for this.

For managing the cluster I am planning on adding a "down" function to be able to terminate it but equally can also just leave that manual management out...


3-enrich/emr-etl-runner/.gitignore
```diff
@@ -21,3 +21,5 @@ deploy/*
 
 # Added as per http://ryan.mcgeary.org/2011/02/09/vendor-everything-still-applies/
 vendor/bundle/ruby
+
+.idea/
```
chuwy: We're avoiding to add environment-specific entries to gitignore.

3-enrich/emr-etl-runner/lib/snowplow-emr-etl-runner/emr.rb
```diff
@@ -0,0 +1,79 @@
...
+      # Parameters:
+      # +client+:: EMR client
+      # +name+:: EMR cluster name
+      def get_emr_jobflow_id(client, name)
+        get_emr_jobflow_id_impl(client, name)
+      end
+
+      private
+
+      def get_emr_jobflow_id_impl(client, name)
```
chuwy: What was the reason for second private function?

3-enrich/emr-etl-runner/Gemfile
```diff
@@ -26,6 +26,7 @@ gem "awrence", "~> 0.1.0"
 gem "snowplow-tracker", "~> 0.5.2"
 gem "iglu-ruby-client", ">= 0.1.0"
 gem "aws-sdk-s3", "~> 1"
+gem "aws-sdk-emr", "~> 1.1.0"
```
chuwy: Elasticity also has the necessary function https://github.com/rslifka/elasticity/blob/master/lib/elasticity/emr.rb#L124-L135.

I generally agree that we should prefer the official SDK rather than wrapper, but in sake of consistency I'd not introduce new dependencies.

3-enrich/emr-etl-runner/lib/snowplow-emr-etl-runner/emr.rb
```diff
@@ -0,0 +1,79 @@
...
+        marker = nil
+        emr_cluster_id = nil
+
+        loop do
+          response = list_clusters(client, marker)
+          emr_clusters = response.clusters.select { |c| c[:name] == name }
+
+          if emr_clusters.size == 0
```
chuwy: I think nested `if` can be nicely replaced with `elsif` of `case` statements.

3-enrich/emr-etl-runner/lib/snowplow-emr-etl-runner/emr.rb
```diff
@@ -0,0 +1,79 @@
...
+      # +client+:: EMR client
+      # +name+:: EMR cluster name
+      def get_emr_jobflow_id(client, name)
+        get_emr_jobflow_id_impl(client, name)
+      end
+
+      private
+
+      def get_emr_jobflow_id_impl(client, name)
+        marker = nil
```
chuwy: Can we add a short note here that `marker` is used for pagination.

3-enrich/emr-etl-runner/.gitignore
jbeemster: Will add it to my global and remove from here - you are right!

3-enrich/emr-etl-runner/lib/snowplow-emr-etl-runner/emr.rb
jbeemster: Following the same code pattern as @BenFradet implemented in his S3 class - happy to drop this back to a single public function.

3-enrich/emr-etl-runner/lib/snowplow-emr-etl-runner/emr.rb
```diff
@@ -0,0 +1,79 @@
...
+              # Cluster be in WAITING for new job steps to be submitted...
+              if emr_cluster[:status][:state] == "RUNNING"
+                raise EmrDiscoveryError, "EMR Cluster must be in WAITING state before new job steps can be submitted - found #{emr_cluster[:status][:state]}"
+              end
+
+              emr_cluster_id = emr_cluster[:id]
+            end
+          end
+
+          break if marker.nil? or !emr_cluster_id.nil?
```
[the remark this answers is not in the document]
jbeemster: At the moment I am ignoring that possibility and it will use the first page that is successful - it didn't seem like a good use of processing power to paginate through every single page of EMR clusters to look for that condition.

I can however change it to first populate a full list of emr_clusters before applying the filtering logic?  Will just be a bit more expensive.

3-enrich/emr-etl-runner/Gemfile.lock
```diff
@@ -137,4 +141,4 @@ RUBY VERSION
    ruby 2.3.1p0 (jruby 9.1.6.0)
 
 BUNDLED WITH
-   1.15.4
+   2.0.0.pre.1
```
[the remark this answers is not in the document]
jbeemster: Fair enough!  Will drop down my local version of bundler back to 1.15.4...

3-enrich/emr-etl-runner/lib/snowplow-emr-etl-runner/errors.rb
```diff
@@ -29,6 +29,10 @@ class ConfigError < Error
     class EmrExecutionError < Error
     end
 
+    # Problem when discovering Amazon EMR JobFlow (e.g. multiple job-flows discovered with the same name)
+    class EmrDiscoveryError < Error
+    end
```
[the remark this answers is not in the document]
jbeemster: We could do - TBH we should never hit that failure in practice as the Consul lock should prevent us from ever attempting to get a RUNNING cluster and we should never have a non-unique cluster.  If you think it would be useful happy to add one in.

stdfalse:
> Also pinging @ihortom @stdfalse for their input on the description above.

The logic looks good to me.

3-enrich/emr-etl-runner/lib/snowplow-emr-etl-runner/emr_job.rb
```diff
@@ -76,6 +78,7 @@ def initialize(debug, staging, enrich, staging_stream_enrich, shred, es, archive
 
         collector_format = config.dig(:collectors, :format)
         run_tstamp = Time.new
+        @run_tstamp = run_tstamp
```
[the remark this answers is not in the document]
BenFradet: I meant combining the two lines into one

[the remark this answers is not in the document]
jbeemster: Doh of course yes we can!

3-enrich/emr-etl-runner/lib/snowplow-emr-etl-runner/emr.rb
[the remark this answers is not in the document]
jbeemster: What would you prefer @BenFradet ?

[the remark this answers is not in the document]
jbeemster: Ended up implementing it to recursively find all clusters which are running or waiting with that name before checking so should be bullet proof now!

3-enrich/emr-etl-runner/Gemfile
jbeemster: Implemented using Elasticity EMR class instead.

jbeemster:
@chuwy @BenFradet I think I have addressed all the main feedback now if you could both have another pass?  Would be great to get this into an RC this week so we can try it out with an internal pipeline.

jbeemster:
One concern I have with the persistence is a note by S3DistCp (https://docs.aws.amazon.com/emr/latest/ReleaseGuide/UsingEMR_s3distcp.html) - namely:

> If S3DistCp is unable to copy some or all of the specified files, the cluster step fails and returns a non-zero error code. If this occurs, S3DistCp does not clean up partially copied files.

Should I be adding a step conditionally to always clean out any partial copies from HDFS?

3-enrich/emr-etl-runner/lib/snowplow-emr-etl-runner/emr.rb
```diff
@@ -0,0 +1,73 @@
...
+          marker = response['Marker'] if response.has_key?('Marker')
+          break if marker.nil?
+        end
+
+        case emr_clusters.size
+        when 0
+          return nil
+        when 1
+          emr_cluster = emr_clusters.first
+          if emr_cluster['Status']['State'] == "RUNNING"
```
BenFradet: is the double indexing safe to do?

jbeemster:
Hey @BenFradet - it might be a good idea to use a more machine readable name for sure.  Historically all of our chatbot actions have used these rather bespoke names without any issues though so I do not think it would be that much of an issue to be fair.

Are you thinking about using a UUID or did you have something else in mind for the name?

jbeemster:
@BenFradet could you link me to the steps required to get an RC of this available in Bintray so we can try it out?

jbeemster:
Hey @BenFradet so after a lot of chatting with AWS the only extra addition I would like to make are some HDFS state maintenance points - namely clearing out HDFS before any new steps are submitted.

This would involve updating your hdfs rmr script to:

1. Use the -skipTrash argument so deleted files are not just moved to trash but are actually deleted
2. Add the ability to check for the existence of a directory before deleting - this lets us run cleanup steps for dirs that might not exist

3-enrich/emr-etl-runner/bin/snowplow-hadoop-fs-rmr-0.2.0.sh
```diff
@@ -0,0 +1,24 @@
...
+# Author::    Joshua Beemster (mailto:support@snowplowanalytics.com)
+# Copyright:: Copyright (c) 2012-2018 Snowplow Analytics Ltd
+# License::   Apache License Version 2.0
+
+# Removes recursively a folder on HDFS
+hadoop fs -test -d $1
+if [ $? == 0 ]; then
+  hadoop fs -rm -r -skipTrash $1
+else
+  echo "Directory $1 does not exist"
```
BenFradet: should we specify a 0 exit code in this case?

3-enrich/emr-etl-runner/lib/snowplow-emr-etl-runner/emr_job.rb
```diff
@@ -1141,11 +1149,18 @@ def self.bootstrap_failure?(jobflow, cluster_step_statuses)
...
-        step.arguments = ["#{bucket}common/emr/snowplow-hadoop-fs-rmr-0.1.0.sh", location]
-        step.name << ": Empty Raw HDFS"
+        step.arguments = ["#{bucket}common/emr/snowplow-hadoop-fs-rmr-0.2.0.sh", location]
+        step.name << ": #{description}"
+        step
+      end
+
+      def get_hdfs_expunge_step
+        step = Elasticity::CustomJarStep.new("command-runner.jar")
+        step.arguments = %W(hdfs dfs -expunge)
```
BenFradet: does this work with `command-runner`? it's not listed in https://docs.aws.amazon.com/emr/latest/ReleaseGuide/emr-commandrunner.html

3-enrich/emr-etl-runner/lib/snowplow-emr-etl-runner/emr_job.rb
```diff
@@ -157,6 +157,14 @@ def initialize(debug, staging, enrich, staging_stream_enrich, shred, es, archive
         csbe = config[:aws][:s3][:buckets][:enriched]
         csbs = config[:aws][:s3][:buckets][:shredded]
 
+        # Clear HDFS if persistent jobflow has been found
+        if found_persistent_jobflow
+          submit_jobflow_step(get_rmr_step(ENRICH_STEP_INPUT, standard_assets_bucket, "Empty Raw HDFS"), use_persistent_jobflow)
+          submit_jobflow_step(get_rmr_step(ENRICH_STEP_OUTPUT, standard_assets_bucket, "Empty Enriched HDFS"), use_persistent_jobflow)
+          submit_jobflow_step(get_rmr_step(SHRED_STEP_OUTPUT, standard_assets_bucket, "Empty Shredded HDFS"), use_persistent_jobflow)
```
BenFradet: could we combine those as one step?

3-enrich/emr-etl-runner/lib/snowplow-emr-etl-runner/emr_job.rb
jbeemster: Tested and working @BenFradet - you can execute any bash command using `command-runner.jar` - the only reason I could not use it for all the removals was that I needed to add if statement checks.

3-enrich/emr-etl-runner/lib/snowplow-emr-etl-runner/emr_job.rb
jbeemster: So the underlying script would accept N HDFS path arguments and loop through them?  Yeah that can be done easily enough.

3-enrich/emr-etl-runner/bin/snowplow-hadoop-fs-rmr-0.2.0.sh
jbeemster: It does it automatically so not sure its required - can add an exit 0 however.

3-enrich/emr-etl-runner/lib/snowplow-emr-etl-runner/emr_job.rb
BenFradet: or just rm `/local/snowplow`

3-enrich/emr-etl-runner/lib/snowplow-emr-etl-runner/cli.rb
```diff
@@ -77,6 +78,7 @@ def self.get_args_config_enrichments_resolver
             opts.on('--ignore-lock-on-start', 'ignore the lock if it is set when starting') { |config| options[:ignore_lock_on_start] = true }
             opts.on('--consul ADDRESS', 'address to the Consul server') { |config| options[:consul] = config }
             opts.on('--use-persistent-jobflow', 'discovers and uses a persistent cluster for steps') { |config| options[:use_persistent_jobflow] = true }
+            opts.on('--persistent-jobflow-duration DURATION', 'the length of time a cluster can for before being replaced') { |config| options[:persistent_jobflow_duration] = config }
```
BenFradet: maybe we should mention the format

3-enrich/emr-etl-runner/lib/snowplow-emr-etl-runner/emr_job.rb
```diff
@@ -719,6 +722,13 @@ def run(config)
           raise EmrExecutionError, get_failure_details(jobflow_id)
         end
 
+        if @use_persistent_jobflow and @persistent_jobflow_duration_s > 0
+          if @jobflow.cluster_status.created_at + @persistent_jobflow_duration_s < @run_tstamp
```
BenFradet: should we check no steps are running?

BenFradet: also can we combine the ifs

ihortom:
While running `r112-rc3` every 2 minutes now and then we come across a failure with the error

```
EMR Cluster must be in WAITING state before new job steps can be submitted - found RUNNING
```

It appears due to starting a new job on the persistent cluster while the cluster has not yet switched its status to `WAITING`. Also, 2 minutes interval is quite short to squeeze in an action like removing the lock. It seems that decreasing run frequency to every 5 mins is more plausible at the moment.

stdfalse:
We have noticed strange behavior on `RestClient::RequestTimeout` issues: in transient mode EMR cluster either not starting or loosing connection and executing all steps, in persistent mode it's executing steps until the disconnect and isn't reporting anything about completed steps. An example:

__stdout__:

```
D, [2019-01-17T03:35:37.280603 #29694] DEBUG -- : Initializing EMR jobflow
```

__stderr__:

```
uri:classloader:/gems/avro-1.8.1/lib/avro/schema.rb:350: warning: constant ::Fixnum is deprecated
uri:classloader:/gems/json-schema-2.7.0/lib/json-schema/util/array_set.rb:18: warning: constant ::Fixnum is deprecated
RestClient::RequestTimeout: Request Timeout
                  transmit at uri:classloader:/gems/rest-client-1.8.0/lib/restclient/request.rb:427
...
```

__last emr status__:

```
| Name                                                                                                 | Status    | Start time                        | Elapsed time     | Elapsed (s) | 
| ---------------------------------------------------------------------------------------------------- | --------- | --------------------------------- | ---------------- | ----------- | 
| Elasticity Spark Step: Shred Enriched Events                                                         | COMPLETED | Thu Jan 17 2019 03:37:57 UTC+0000 |  0 h,  1 m, 56 s | 116         | 
| Elasticity S3DistCp Step: Enriched S3 -> HDFS                                                        | COMPLETED | Thu Jan 17 2019 03:37:27 UTC+0000 |  0 h,  0 m, 28 s | 28          | 
| Elasticity S3DistCp Step: Stream Enriched s3a://xxx-kinesis-s3-enriched/main/ -> Enriched Staging S3 | COMPLETED | Thu Jan 17 2019 03:36:55 UTC+0000 |  0 h,  0 m, 30 s | 30          | 
| Elasticity Custom Jar Step: Empty HDFS trash                                                         | COMPLETED | Thu Jan 17 2019 03:36:47 UTC+0000 |  0 h,  0 m,  6 s | 6           | 
| Elasticity Custom Jar Step: Empty Snowplow HDFS                                                      | COMPLETED | Thu Jan 17 2019 03:36:27 UTC+0000 |  0 h,  0 m, 18 s | 18          | 
```

As you can see, it has just stopped after shred step and no indication about this in logs. Is it possible to fetch this info or make it to execute all steps so it won't fail in the middle of the way?

BenFradet:
hey @stdfalse , the connection isn't maintained for the duration of the steps, just when sending steps to be executed, can you link the full incident?

BenFradet:
incorporated into #3945

PR #3941, opened 2018-12-17
Stream enrich: Extend config to override kafka producer and consumer config

0xE282B0:
Based on PR https://github.com/snowplow/snowplow/pull/3938 this PR extends the configuration of [stream-enrich](https://github.com/snowplow/snowplow/tree/master/3-enrich/stream-enrich)  to override the configuration of the kafka producer and consumer.

### DIFF
--- a/3-enrich/stream-enrich/core/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/SpecHelpers.scala
+++ b/3-enrich/stream-enrich/core/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/SpecHelpers.scala
@@ -157,7 +157,7 @@ object SpecHelpers {
       streams = StreamsConfig(
         InConfig("raw"),
         OutConfig("enriched", Some("pii"), "bad", "partitionkey"),
-        Kafka("brokers", 1, None),
+        Kafka("brokers", 1, None, None),
         BufferConfig(1000L, 100L, 1200L),
         "appName"
       ),
--- a/3-enrich/stream-enrich/core/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/SpecHelpers.scala
+++ b/3-enrich/stream-enrich/core/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/SpecHelpers.scala
@@ -157,7 +157,7 @@ object SpecHelpers {
       streams = StreamsConfig(
         InConfig("raw"),
         OutConfig("enriched", Some("pii"), "bad", "partitionkey"),
-        Kafka("brokers", 1),
+        Kafka("brokers", 1, None),
         BufferConfig(1000L, 100L, 1200L),
         "appName"
       ),
--- a/3-enrich/stream-enrich/examples/config.hocon.sample
+++ b/3-enrich/stream-enrich/examples/config.hocon.sample
@@ -96,6 +96,18 @@ enrich {
       brokers = "{{kafkaBrokers}}"
       # Number of retries to perform before giving up on sending a record
       retries = 0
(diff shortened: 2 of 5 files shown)


snowplowcla:
Confirmed! @0xE282B0  has signed the Individual Contributor License Agreement. Thanks so much

3-enrich/stream-enrich/core/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/SpecHelpers.scala
```diff
@@ -157,7 +157,7 @@ object SpecHelpers {
       streams = StreamsConfig(
         InConfig("raw"),
         OutConfig("enriched", Some("pii"), "bad", "partitionkey"),
-        Kafka("brokers", 1),
+        Kafka("brokers", 1, null, null),
```
BenFradet: Could you replace the nulls with nones?

3-enrich/stream-enrich/examples/config.hocon.sample
```diff
@@ -96,6 +96,17 @@ enrich {
...
       retries = 0
+      # The kafka producer has a variety of possible configuration options defined at
+      # https://kafka.apache.org/documentation/#producerconfigs
+      #producerConf {
+      #  "compression.type" = snappy
+      #}
+      # The kafka consumer has a variety of possible configuration options defined at
+      # https://kafka.apache.org/documentation/#consumerconfigs
+      #consumerConf {
+      #  "compression.type" = snappy
```
BenFradet: compression.type is not a consumer configuration

3-enrich/stream-enrich/examples/config.hocon.sample
```diff
@@ -96,6 +96,17 @@ enrich {
       brokers = "{{kafkaBrokers}}"
       # Number of retries to perform before giving up on sending a record
       retries = 0
+      # The kafka producer has a variety of possible configuration options defined at
+      # https://kafka.apache.org/documentation/#producerconfigs
+      #producerConf {
```
BenFradet: we might want to specify the configuration that are already used so that there is no override, wdyt?

0xE282B0: Good point. However, since some values are set from other values in the configuration file, I would suggest only specifying the remaining (fixed) values and referring to the other locations.
Like this: https://gist.github.com/0xE282B0/d91b8fac239c360cfd4c8e58901f87b0#file-config-hocon-sample-L94

Did I get you right?

BenFradet: I'm not sure I fully understand what you mean, could you give an example? either here or as an additional commit :+1:

0xE282B0: I've added all values that are hardcoded in the `createProperties(...)` methods of the [source](https://github.com/snowplow/snowplow/blob/master/3-enrich/stream-enrich/kafka/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/sources/KafkaSource.scala#L125) and [sink](https://github.com/snowplow/snowplow/blob/master/3-enrich/stream-enrich/kafka/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/sinks/KafkaSink.scala#L53).

For the other values which are collected from other parts of the config I added a comment. Maybe it's a bit to crowdy.

0xE282B0:
Unfortunately, Travis could not build. Since I could not reproduce it, I tried to trigger a rebuild and accidentally deleted the branch. 😆

BenFradet:

PR #3954, opened 2019-01-30
Allow to override config settings with ENV variables.

christoph-buente:
In order to configure the collector and enricher using environment variables in AWS Beanstalk, ECS or even kubernetes with config maps instead of a fixed config file, we augmented the hocon sample files with additional (and conditional) evaluation of env variables. We used a naming scheme for transforming the nested hocon keys into env variables:
Each nested level will be separated with an underscore and appended. Also Camel case is turned separate words joined with underscore. All characters are upper cased then.
    
    collector {
      crossDomain {
        enabled
      }
    }

becomes:
`COLLECTOR_CROSS_DOMAIN_ENABLED`

I hope this PR will make the adoption of snowplow on docker much easier. 
    
    docker run \
      -v $PWD/:/snowplow/config \
      -e COLLECTOR_COOKIE_DOMAIN="your-domain" \
      -e COLLECTOR_COOKIE_NAME="custom-cookie-name" \
      -e COLLECTOR_COOKIE_EXPIRATION="730 days" \
      -e COLLECTOR_PORT="8091" \
      -e GOOD_STREAM="good-stream-name" \
      -e BAD_STREAM="bad-stream-name" \
      -e MAX_BACKOFF_MILLIS=3000 \
      -e MIN_BACKOFF_MILLIS=600000 \
      -e AWS_REGION="us-east-1" \
      -e BUFFER_RECORD_THRESHOLD=100 \
      -e BUFFER_TIME_THRESHOLD=1000 \
      -e BUFFER_BYTE_THRESHOLD=102400 \
      -e AWS_ACCESS_KEY_ID=KEY \
      -e AWS_SECRET_ACCESS_KEY=SECRET \
      snowplow-docker-registry.bintray.io/snowplow/scala-stream-collector-kinesis:0.14.0 \
      --config /snowplow/config/config.hocon

### DIFF
--- a/3-enrich/stream-enrich/examples/config.hocon.sample
+++ b/3-enrich/stream-enrich/examples/config.hocon.sample
@@ -20,16 +20,20 @@ enrich {
 
     in {
       # Stream/topic where the raw events to be enriched are located
-      raw = {{streamsInRaw}}
+      raw = "{{streamsInRaw}}"
+      raw = ${?ENRICH_STREAMS_IN_RAW}
     }
 
     out {
       # Stream/topic where the events that were successfully enriched will end up
-      enriched = {{outEnriched}}
+      enriched = "{{outEnriched}}"
+      enriched = ${?ENRICH_STREAMS_OUT_ENRICHED}
       # Stream/topic where the event that failed enrichment will be stored
-      bad = {{outBad}}
+      bad = "{{outBad}}"
+      bad = ${?ENRICH_STREAMS_OUT_BAD}
       # Stream/topic where the pii tranformation events will end up
-      pii = {{outPii}}
+      pii = "{{outPii}}"
+      pii = ${?ENRICH_STREAMS_OUT_PII}
 
       # How the output stream/topic will be partitioned.
       # Possible partition keys are: event_id, event_fingerprint, domain_userid, network_userid,
@@ -38,7 +42,8 @@ enrich {
       # possible parittion keys correspond to.
       # Otherwise, the partition key will be a random UUID.
       # Note: Nsq does not make use of partition key.
-      partitionKey = {{partitionKeyName}}
+      partitionKey = "{{partitionKeyName}}"
+      partitionKey = ${?ENRICH_STREAMS_OUT_PARTITION_KEY}
     }
 
     # Configuration shown is for Kafka, to use another uncomment the appropriate configuration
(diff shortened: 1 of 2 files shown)


christoph-buente:
Looks like the changes break the test setup. I apologize, but i cannot run the tests locally. Maybe @BenFradet can take a look and say, wether this PR is considered useful and how to fix the tests.

christoph-buente:
Technically, it is not limited to docker. You could also run the jar on bare metal and use environment variable for configuration rather than fixed file. This falls into the domain of 12 factor apps: https://12factor.net/

That being said, i looked into the snowplow-docker repo. It looks like for each release version and each stream provider you need a different config file, which is not part of the repo yet. That's why i thought it would be better to have a working base version of the config file in the actual code repo available to have it shipped with the docker images.

What do you think @BenFradet?

BenFradet:
>That's why i thought it would be better to have a working base version of the config file in the actual code repo available to have it shipped with the docker images.

right, agreed :+1: , regarding the tests failing could you leave the quoting as it was? there are now strings that should be numbers

christoph-buente:
I can certainly revert the double quote changes. However, the hocon parser fails to interpret the {{placeholders}} in the file. What mechanism is being used to replace those placeholders with real values? It seems to m,0e there is another replacement process, similar to the environment variables.

Is it possible to have real values in the file with sensible defaults?

christoph-buente:
I removed the quoting, but i checked the hocon format documentation. And it looks like unquoted `{` and `}` characters are actually forbidden. @BenFradet How do you make the hocon parser work with it? I did not succeed.

3-enrich/stream-enrich/examples/config.hocon.sample
```diff
@@ -53,61 +58,71 @@ enrich {
...
       # (pertinent to kinesis source type)
       # initialPosition = TRIM_HORIZON
+      # initialPosition = ${?ENRICH_STREAMS_SOURCE_SINK_INITIAL_POSITION}
 
       # Need to be specified when initial-position is "AT_TIMESTAMP".
       # Timestamp format need to be in "yyyy-MM-ddTHH:mm:ssZ".
       # Ex: "2017-05-17T10:00:00Z"
       # Note: Time need to specified in UTC.
-      initialTimestamp = "{{initialTimestamp}}"
+      initialTimestamp = {{initialTimestamp}}
```
BenFradet: this had quotes originally, that's why the tests are failing :+1: 

If you're interested:

- templating goes on here: https://github.com/snowplow/snowplow/blob/master/3-enrich/stream-enrich/integration-tests/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/PiiEmitSpec.scala#L77
- the example config is turned into a resource here: https://github.com/snowplow/snowplow/blob/master/3-enrich/stream-enrich/project/BuildSettings.scala#L56-L58

BenFradet: it's true for other places too, some quotes have been unnecessarily remove

snowplowcla:
@christoph-buente  has signed the Software Grant and Corporate Contributor License Agreement

BenFradet:

PR #3956, opened 2019-02-01
Spark Enrich: bump Spark to 2.3.0 (closes #3761)

### LINKED ISSUE #3761: Spark Enrich: bump Spark to 2.4.4
needs to happen at the same time as snowplow/snowplow-rdb-loader#101

arunma:

### DIFF
--- a/3-enrich/spark-enrich/project/Dependencies.scala
+++ b/3-enrich/spark-enrich/project/Dependencies.scala
@@ -32,7 +32,7 @@ object Dependencies {
     val elephantBird     = "4.14"
     val geoip2           = "2.5.0"
     // Scala
-    val spark            = "2.2.0"
+    val spark            = "2.3.2"
     val scalaz7          = "7.0.9"
     val scopt            = "3.5.0"
     val commonEnrich     = "0.35.0"
--- a/3-enrich/spark-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.spark/EnrichJob.scala
+++ b/3-enrich/spark-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich.spark/EnrichJob.scala
@@ -74,7 +74,9 @@ object EnrichJob extends SparkJob {
     classOf[scala.collection.immutable.Set$EmptySet$],
     classOf[scala.collection.mutable.WrappedArray$ofRef],
     classOf[org.apache.spark.internal.io.FileCommitProtocol$TaskCommitMessage],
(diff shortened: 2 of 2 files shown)


arunma:
@BenFradet Tried upgrading to Spark to 2.4.0, as suggested. Bumped into two issues and I would like to know your thoughts on this. 

1. `Caused by: com.fasterxml.jackson.databind.JsonMappingException: Incompatible Jackson version: 2.8.4`

Managed to get through this issue by adding the following Overrides

**build.sbt**
```
.settings(
    dependencyOverrides ++= Seq(
      Dependencies.Overrides.jacksonCore,
      Dependencies.Overrides.jacksonDatabind,
      Dependencies.Overrides.jacksonModuleScala,
      Dependencies.Overrides.json4s
    )
  )
```

**Dependencies.scala**

```
 object Overrides {
    val jacksonCore = "com.fasterxml.jackson.core" % "jackson-core" % "2.8.4"
    val jacksonDatabind = "com.fasterxml.jackson.core" % "jackson-databind" % "2.8.4"
    val jacksonModuleScala = "com.fasterxml.jackson.module" %% "jackson-module-scala" % "2.8.4"
    val json4s = "org.json4s" %% "json4s-jackson" % "3.2.11"
  }
```
However, I am not convinced if this is the best way to approach this. Can't google a better option too.

2. After the above overrides, a couple of testcases fail - `CljTomcatPingdomEventSpec` and `CljTomcatPagerdutyEventSpec`, I notice that with Spark 2.4, the results are empty string ("") instead of the expected `null`s. I am not familiar with the code yet and I would appreciate your inputs on this. 

```
val expected = List(
    "uptime",
    "srv",
    etlTimestamp,
    "2014-10-09 16:28:31.000",
    null,
    "unstruct",
    null, // We can't predict the event_id
    null,
```
On debug window

```
actual = {String[131]@12670} 
 0 = "uptime"
 1 = "srv"
 2 = "2001-09-09 01:46:40.000"
 3 = "2014-10-09 16:28:31.000"
 4 = """"
 5 = "unstruct"
 6 = "30d288f6-b2bf-44db-9fe8-7872ae3d4713"
 7 = """"
 8 = """"
```

BenFradet:
ah the infamous jackson...

I think we should try aligning to Spark, which uses 2.6.7.1: https://mvnrepository.com/artifact/org.apache.spark/spark-core_2.12/2.4.0.

And list/check which dependencies are incompatible.

I realize this is more involved than the original ticket, so feel free to leave that to me, we'll have to do it in the near future anyway :+1:

chuwy:
Closing in favor of https://github.com/snowplow/snowplow/issues/3761

Issue #3880, opened 2018-08-24
Scala Common Enrich: explore more relaxed URL parsing


BenFradet:
especially wrt macros (${a}, #{a}, %%a%%)

PR #4003, opened 2019-03-04
[SCE] Relaxed url parsing (issue #3880)

benjben:
Use [scala-uri](https://github.com/lemonlabsuk/scala-uri) by default, which is way more relaxed than `java.net.URI`.
Correctly percent-encoded URIs are then passed to `java.net.URI`.
If a URI is expected to contain macros with %, `percentDecode` should be set to `false` when calling the function `stringToUri`.

### DIFF
--- a/3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/ConversionUtils.scala
+++ b/3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/ConversionUtils.scala
@@ -31,9 +31,6 @@ import scala.util.control.NonFatal
 // Apache HTTP
 import org.apache.http.client.utils.URLEncodedUtils
 
-// Apache Commons
-import org.apache.commons.lang3.exception.ExceptionUtils
-
 // Apache Commons Codec
 import org.apache.commons.codec.binary.Base64
 
@@ -42,7 +39,11 @@ import scalaz._
 import Scalaz._
 
 // Scala URI
-import com.netaporter.uri.Uri
+import io.lemonlabs.uri.Uri
+import io.lemonlabs.uri.Url
+import io.lemonlabs.uri.config.UriConfig
+import io.lemonlabs.uri.decoding.PercentDecoder
+import io.lemonlabs.uri.encoding.percentEncode
 
 /**
  * General-purpose utils to help the
@@ -311,52 +312,35 @@ object ConversionUtils {
     URLEncoder.encode(str, enc)
 
   /**
-   * A wrapper around Java's
-   * URI.create().
+   * Parses a string to create a [[URI]].
+   * Parsing is relaxed, i.e. even if a URL is not correctly percent-encoded or not RFC 3986-compliant, it can be parsed.
    *
-   * Exceptions thrown by
-   * URI.create():
-   * 1. NullPointerException
-   *    if uri is null
-   * 2. IllegalArgumentException
-   *    if uri violates RFC 2396
-   *
-   * @param uri The URI string to
-   *        convert
-   * @param useNetaporter Whether to use the
-   *        com.netaporter.uri library
-   * @return an Option-boxed URI object, or an
-   *         error message, all
-   *         wrapped in a Validation
+   * @param uri String containing the URI to parse.
+   * @return [[Validation]] wrapping the result of the parsing:
+   *         - [[Success]] with the parsed URI if there was no error or with [[None]] if the input was `null`.
+   *         - [[Failure]] with the error message if something went wrong.
    */
-  def stringToUri(uri: String, useNetaporter: Boolean = false): Validation[String, Option[URI]] =
-    try {
-      val r = uri.replaceAll(" ", "%20") // Because so many raw URIs are bad, #346
-      Some(URI.create(r)).success
-    } catch {
-      case npe: NullPointerException => None.success
-      case iae: IllegalArgumentException =>
-        useNetaporter match {
-          case false => {
-            val netaporterUri = try {
-              Uri.parse(uri).success
-            } catch {
-              case NonFatal(e) =>
-                "Provided URI string [%s] could not be parsed by Netaporter: [%s]"
-                  .format(uri, ExceptionUtils.getRootCause(iae).getMessage)
-                  .fail
-            }
-            for {
-              parsedUri <- netaporterUri
-              finalUri  <- stringToUri(parsedUri.toString, true)
-            } yield finalUri
-          }
-          case true =>
-            "Provided URI string [%s] violates RFC 2396: [%s]"
-              .format(uri, ExceptionUtils.getRootCause(iae).getMessage)
-              .fail
+  def stringToUri(uri: String): Validation[String, Option[URI]] =
+    Try(
+      Option(uri) // to handle null
+        .map(_.replaceAll(" ", "%20"))
+        .map(URI.create)
+    ) match {
+      case util.Success(parsed) =>
+        parsed.success
+      case util.Failure(javaErr) =>
+        implicit val c =
+          UriConfig(decoder = PercentDecoder(ignoreInvalidPercentEncoding = true), encoder = percentEncode -- '+')
+        Uri
+          .parseTry(uri)
+          .map(_.toJavaURI) match {
+          case util.Success(javaURI) =>
+            Some(javaURI).success
+          case util.Failure(scalaErr) =>
+            "Provided URI [%s] could not be parsed, neither by Java parsing (error: [%s]) nor by Scala parsing (error: [%s])."
+              .format(uri, javaErr.getMessage, scalaErr.getMessage)
+              .failure
         }
-      case NonFatal(e) => "Unexpected error creating URI from string [%s]: [%s]".format(uri, e.getMessage).fail
     }
 
   /**
@@ -368,7 +352,7 @@ object ConversionUtils {
   def extractQuerystring(uri: URI, encoding: String): Validation[String, Map[String, String]] =
     Try(URLEncodedUtils.parse(uri, encoding).map(p => (p.getName -> p.getValue))).recoverWith {
       case NonFatal(_) =>
-        Try(Uri.parse(uri.toString).query.params).map(l => l.map(t => (t._1, t._2.getOrElse(""))))
+        Try(Url.parse(uri.toString).query.params).map(l => l.map(t => (t._1, t._2.getOrElse(""))))
     } match {
       case util.Success(s) => s.toMap.success
       case util.Failure(e) => s"Could not parse uri [$uri]. Uri parsing threw exception: [$e].".fail
--- a/3-enrich/scala-common-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/registry/EnrichmentConfigsSpec.scala
+++ b/3-enrich/scala-common-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/registry/EnrichmentConfigsSpec.scala
@@ -458,7 +458,7 @@ class EnrichmentConfigsSpec extends Specification with ValidationMatchers with D
           },
           "includeUseragentFile": {
              "database": "include_current.txt",
-             "uri": "invalid\\uri"
+             "uri": "file://foo:{waaat}/"
           }
         }
       }""")
--- a/3-enrich/scala-common-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/utils/conversionUtilsSpecs.scala
+++ b/3-enrich/scala-common-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/utils/conversionUtilsSpecs.scala
@@ -22,38 +22,109 @@ import Scalaz._
 
 // Specs2
 import org.specs2.{ScalaCheck, Specification}
+import org.specs2.mutable.{Specification => MutableSpecification}
 import org.specs2.matcher.DataTables
 import org.specs2.scalaz.ValidationMatchers
-import org.scalacheck._
-import org.scalacheck.Arbitrary._
 
-class StringToUriSpec extends Specification with DataTables {
+class StringToUriSpec extends MutableSpecification with ValidationMatchers {
 
-  def is = s2"Parsing Strings into URIs should work $e1"
+  /** Helper to generate URLs with `chars` at different places in the path and in the query string, doubled, tripled, etc. */
+  private def generateUrlsWithChars(chars: String): List[String] = List(
+    s"http://www.example.com/a/b/$chars",
+    s"http://www.example.com/a$chars",
+    s"http://www.example.com/a$chars$chars",
+    s"http://www.example.com/a$chars$chars${chars}b/c",
+    s"http://www.example.com/a${chars}/c$chars${chars}d",
+    s"http://www.example.com/a${chars}b/456?d=e$chars${chars}f&g=h${chars}i&j=k",
+    s"http://www.example.com/a${chars}b/c?d=e&f=g$chars$chars${chars}h"
+  )
 
-  def e1 =
-    "SPEC NAME"     || "URI" | "EXPECTED" |
-      "Empty URI"   !! null ! None.success |
-      "Simple URI"  !! "https://google.com" ! Some(URI.create("https://google.com")).success |
-      "Complex URI" !! "http://www.google.com/search?q=gateway+oracle+cards+denise+linn&hl=en&client=safari" ! Some(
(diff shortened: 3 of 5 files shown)


3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/ConversionUtils.scala
```diff
@@ -311,52 +311,44 @@ object ConversionUtils {
...
+      case e: Exception =>
+        relaxed match {
+          case false =>
+            "Provided URI [%s] could not be parsed (without relaxed parsing). Error: [%s]"
+              .format(uri, e.getMessage)
+              .fail
+          case true =>
+            try {
+              implicit val config = UriConfig(decoder = NoopDecoder) // no percent-decoding
+              Option(uri).map(Uri.parse).map(_.toJavaURI).success
```
BenFradet: there is https://github.com/lemonlabsuk/scala-uri/blob/07756b8e2eb7de1318767edbc730cee6b768b736/shared/src/main/scala/io/lemonlabs/uri/Uri.scala#L556-L560

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/ConversionUtils.scala
```diff
@@ -311,52 +311,44 @@ object ConversionUtils {
...
-      Some(URI.create(r)).success
+      Option(uri).map(_.replaceAll(" ", "%20")).map(URI.create).success
     } catch {
-      case npe: NullPointerException => None.success
-      case iae: IllegalArgumentException =>
-        useNetaporter match {
-          case false => {
-            val netaporterUri = try {
-              Uri.parse(uri).success
+      case e: Exception =>
```
BenFradet: this catches everything and should be avoided

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/ConversionUtils.scala
```diff
@@ -311,52 +311,44 @@ object ConversionUtils {
...
+              .fail
+          case true =>
+            try {
+              implicit val config = UriConfig(decoder = NoopDecoder) // no percent-decoding
+              Option(uri).map(Uri.parse).map(_.toJavaURI).success
             } catch {
-              case NonFatal(e) =>
-                "Provided URI string [%s] could not be parsed by Netaporter: [%s]"
-                  .format(uri, ExceptionUtils.getRootCause(iae).getMessage)
+              case e: Exception =>
```
BenFradet: same

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/ConversionUtils.scala
```diff
@@ -311,52 +311,44 @@ object ConversionUtils {
...
+   *
+   * @param uri String containing the URI to parse.
+   * @param relaxed Specifies whether parsing should be relaxed or not.
+   *
+   * @return [[Validation]] wrapping the result of the parsing:
+   *         - [[Success]] with the parsed URI if there was no error or with [[None]] if the input was `null`.
+   *         - [[Failure]] with the error message if something went wrong.
    */
-  def stringToUri(uri: String, useNetaporter: Boolean = false): Validation[String, Option[URI]] =
+  def stringToUri(uri: String, relaxed: Boolean = false): Validation[String, Option[URI]] =
```
BenFradet: should we ditch `relaxed` and use scala-uri directly?

3-enrich/scala-common-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/utils/conversionUtilsSpecs.scala
```diff
@@ -22,38 +22,90 @@ import Scalaz._
...
+        "http://www.example.com",
+        "http://www.example.com/",
+        "http://www.example.com/a",
+        "http://www.google.com/se+arch?q=gateway+oracle+cards+denise+linn&hl=en&client=safari"
+      ) ++ generateUrlsWithChars(""))
+        .map(url => ConversionUtils.stringToUri(url) must_== Some(URI.create(url)).success)
+    }
+
+    "work with URL with space" >> {
+      ConversionUtils.stringToUri("http://www.example.com/sp ace") must beSuccessful
```
BenFradet: can we check the result?

3-enrich/scala-common-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/utils/conversionUtilsSpecs.scala
```diff
@@ -22,38 +22,90 @@ import Scalaz._
...
+        generateUrlsWithChars("#a#") ++
+        generateUrlsWithChars("#a b#") ++
+        generateUrlsWithChars("##a##") ++
+        generateUrlsWithChars("##a b##") ++
+        generateUrlsWithChars("%a%") ++
+        generateUrlsWithChars("%a b%}") ++
+        generateUrlsWithChars("%%a%%}") ++
+        generateUrlsWithChars("%%a b%%}") ++
+        generateUrlsWithChars("%%%a%%%") ++
+        generateUrlsWithChars("%%%a b%%%")
```
BenFradet: would we be able to verify their result as well?

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/ConversionUtils.scala
benjben: Don't we want the function never to throw an exception, but instead return the error message is something went wrong?

BenFradet: yes but this catches OOM, InterruptedException, etc which isn't what we want

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/ConversionUtils.scala
benjben: I hesitated about it, but I think that it should be left as a possibility to do a strict URI parsing. But we can have a relaxed URI parsing by default and make it strict optionally, instead of the other way around. WDYT?

For example see [this](https://github.com/snowplow/snowplow/blob/master/3-enrich/scala-common-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/registry/EnrichmentConfigsSpec.scala#L446)

BenFradet: makes sense, in this case we could split the two, one that uses java and is strict and one that uses scala-uri for non-strict usages

benjben: Actually, the choice of relaxed parsing or not would still need to go all the way from the configuration to the function:
```
$ grep -R stringToUri
enrichments/EnrichmentManager.scala:    val refererUri = CU.stringToUri(event.page_referrer)
enrichments/registry/IpLookupsEnrichment.scala:      .stringToUri(uri + "/" + database)
enrichments/registry/IabEnrichment.scala:      .stringToUri(uri + (if (uri.endsWith("/")) "" else "/") + database)
enrichments/registry/UaParserEnrichment.scala:      .stringToUri(uri + (if (uri.endsWith("/")) "" else "/") + database)
enrichments/web/PageEnrichments.scala:      case (Some(r), None)    => CU.stringToUri(r)
enrichments/web/PageEnrichments.scala:      case (None, Some(t))    => CU.stringToUri(t)
enrichments/web/PageEnrichments.scala:      case (Some(r), Some(t)) => CU.stringToUri(t) // Tracker URL takes precedence
utils/ConversionUtils.scala:  def stringToUri(uri: String, relaxed: Boolean = true): Validation[String, Option[URI]] =
```
Although it's technically possible, I'm not sure that this is worth it to make it happen. Can you think of any case where we wouldn't want a relaxed URI parsing? If not I think that we should just always use relaxed parsing, as you suggested.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/ConversionUtils.scala
```diff
@@ -311,43 +311,34 @@ object ConversionUtils {
...
       Option(uri).map(_.replaceAll(" ", "%20")).map(URI.create).success
     } catch {
-      case e: Exception =>
-        relaxed match {
-          case false =>
-            "Provided URI [%s] could not be parsed (without relaxed parsing). Error: [%s]"
+      case NonFatal(e) =>
+        try {
+          implicit val c = UriConfig(decoder = PercentDecoder(ignoreInvalidPercentEncoding = true))
+          Uri.parseOption(uri).map(_.toJavaURI).success
```
BenFradet: I'd rather do:

```scala
for {
  scalaUri <- Uri.parseTry(uri).toEither
  uri <- Either.catchNonFatal(scalaUri.toJavaURI)
} yield uri
```

without the try/catch.

Also I'd remove the usual URI parsing above

benjben: Neat! But are you referring to
1) Scala 2.12's `Try.toEither` ?
2) cats's `Either.catchNonFatal` ? (scalaz used here)

Ok I will remove java URI parsing then. (I also created this [issue](https://github.com/lemonlabsuk/scala-uri/issues/30) regarding the fact that `scala-uri` re-encodes already correctly percent-encoded URIs.)

BenFradet: yes, there are equivalent with scalaz `Validation` for the second and you can do the first manually easily

benjben: I can't have something as neat because:
1) null url needs to be taken care of before `parseTry`
2) a `String` is returned, and not a `Throwable`

benjben:
The original plan was to use only scala-uri for parsing (in a very relaxed way) and not `java.net.URI.create()` any more.
Although this was working well, a consequence was that some URIs were encoded differently as before, thus breaking unit tests in jobs using SCE.
It's possible to fix the unit tests but they probably don't cover all the cases that we have in our customers' pipelines and updating the output of a function that is used for every single event in all the pipeline is risky.
So it was decided to continue to use Java parsing first and in case of failure a very relaxed Scala parsing (no percent decoding).

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/ConversionUtils.scala
```diff
@@ -42,7 +39,11 @@ import scalaz._
 import Scalaz._
 
 // Scala URI
-import com.netaporter.uri.Uri
+import io.lemonlabs.uri.Uri
+import io.lemonlabs.uri.Url
```
BenFradet: double import

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/ConversionUtils.scala
```diff
@@ -311,53 +312,37 @@ object ConversionUtils {
...
+          .transform( // creates Try[Validation[String, Option[URI]]]. Try will always be a Success
+            s => util.Success(Some(s).success),
+            scalaErr =>
+              util.Success(
+                "Provided URI [%s] could not be parsed, neither by Java parsing (error: [%s]) nor by Scala parsing (error: [%s])."
+                  .format(uri, javaErr.getMessage, scalaErr.getMessage)
+                  .failure
+            )
+          )
+    }.get
```
BenFradet: avoid `get`

3-enrich/scala-common-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/registry/EnrichmentConfigsSpec.scala
```diff
@@ -457,8 +457,7 @@ class EnrichmentConfigsSpec extends Specification with ValidationMatchers with D
             "uri": "https://example.com"
           },
           "includeUseragentFile": {
-             "database": "include_current.txt",
-             "uri": "invalid\\uri"
+             "database": "include_current.txt"
```
BenFradet: can't we create an invalid uri?

BenFradet: if not we need to remove this test

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/ConversionUtils.scala
benjben: There is once Ur*i* and once Ur*l*

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/ConversionUtils.scala
benjben: I was 100% sure that it was a `Success`. I replaced by pattern matching

3-enrich/scala-common-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/registry/EnrichmentConfigsSpec.scala
benjben: It was not so easy to find an invalid one with the relaxed parsing

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/ConversionUtils.scala
```diff
@@ -311,52 +312,35 @@ object ConversionUtils {
...
+    ) match {
+      case util.Success(parsed) =>
+        parsed.success
+      case util.Failure(javaErr) =>
+        implicit val c =
+          UriConfig(decoder = PercentDecoder(ignoreInvalidPercentEncoding = true), encoder = percentEncode -- '+')
+        Uri
+          .parseTry(uri)
+          .map(_.toJavaURI) match {
+          case util.Success(javaURI) =>
```
BenFradet: :art: indent

3-enrich/scala-common-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/registry/EnrichmentConfigsSpec.scala
```diff
@@ -458,7 +458,7 @@ class EnrichmentConfigsSpec extends Specification with ValidationMatchers with D
           },
           "includeUseragentFile": {
              "database": "include_current.txt",
-             "uri": "invalid\\uri"
+              "uri": "file://foo:{waaat}/"
```
BenFradet: :art: indent

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/ConversionUtils.scala
benjben: Arf, `scalafmt` did that :/

benjben:
Superseded by #4076

PR #4012, opened 2019-03-10
SOPHI-532: HTTP Remote Adapter (closes #3760)

### LINKED ISSUE #3760: Scala Common Enrich: add HTTP remote adapter
I was thinking about offloading some of the processes that the Adapter does, via Akka to the other machines. This might help a lot for performance and scalability of stream enrich component when handling different amount of traffic during the day.

My design would be like this:
- Hocon config file:  It would have a setting for RemoteAdapters, with the vendor name.
- AdapterRegistry: It would receive the setting from config file for the new vendor names.
- RemoteAdaptor: This is the new class that will extend **Adapter**.  Its **toRawEvents** happens in the remote Akka Actor and it returns the result when ready.

What do you folks think?

szareiangm:
- [x] Rebasing with master and latest changes
- [ ] Adapting their last comment: sending text stream and receiving it.

### DIFF
--- a/3-enrich/remote-adapter-example/.gitignore
+++ b/3-enrich/remote-adapter-example/.gitignore
@@ -1,0 +1,11 @@
+*.class
+*.log
+
+# sbt specific
+dist/*
+target/
+lib_managed/
+src_managed/
+project/boot/
+project/plugins/project/
+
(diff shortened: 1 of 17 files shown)


szareiangm:
Sorry, wrong repo :(

PR #4023, opened 2019-03-26
Scala Common Enrich: Support multiple instances of API request enrichment

szareiangm:
This is the draft of what I have done so far. Honestly, making this code compile with the changes was not straight-forward. So, I thought it might be better to have everybody's opinions about the code design taken into account from the beginning. 

The change in the nutshell is that we return `List[Enrichment]` instead of `Enrichment` wherever needed.

After the design discussions, I will address the comments and will adjust the unit-tests.  

Please take a look at the changes and tell what do you think.

CC: @Nafid @dmatth


benjben:
Closing as enrich got moved to its [new home](https://github.com/snowplow/enrich). @szareiangm if you still need this feature, please open a PR there.

FYI we started to write a version of enrich with [fs2](https://github.com/typelevel/fs2). If the tests are successful, we might consider rewriting Common Enrich with it. If we do, we will see to support multiple instances of enrichments and their dynamic chaining.

PR #4047, opened 2019-04-25
SCE: X-Forwarded-For takes priority over Forwarded: for= and remove port

benjben:


3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/loaders/IpAddressExtractor.scala
```diff
@@ -30,20 +31,29 @@ object IpAddressExtractor {
...
     headers match {
       case h :: t =>
         h.toLowerCase match {
-          case XForwardedForRegex(originalIpAddress) => originalIpAddress
-          case ForwardedForRegex(originalIpAddress)  => originalIpAddress
-          case _                                     => extractIpAddress(t, lastIp)
+          case XForwardedForRegex(ipv4, ipv6) => Option(ipv4).getOrElse(ipv6)
+          case ForwardedForRegex(ipv4, ipv6) =>
+            val ip = Option(ipv4).getOrElse(ipv6)
+            extractIpAddress(t, lastIp, Some(ip))
```
BenFradet: this should be `Option(ip)` no? can the two groups be null?

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/loaders/IpAddressExtractor.scala
```diff
@@ -30,20 +31,29 @@ object IpAddressExtractor {
...
   @tailrec
-  def extractIpAddress(headers: List[String], lastIp: String): String =
+  def extractIpAddress(headers: List[String], lastIp: String, maybeForwardedForIp: Option[String] = None): String =
     headers match {
       case h :: t =>
         h.toLowerCase match {
-          case XForwardedForRegex(originalIpAddress) => originalIpAddress
-          case ForwardedForRegex(originalIpAddress)  => originalIpAddress
-          case _                                     => extractIpAddress(t, lastIp)
+          case XForwardedForRegex(ipv4, ipv6) => Option(ipv4).getOrElse(ipv6)
```
BenFradet: same, can the two groups be null?

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/loaders/IpAddressExtractor.scala
benjben: If the 2 groups were null the regex would not pattern match.
```
scala> val Regex = """(a{2})|(a{3})""".r
Regex: scala.util.matching.Regex = (a{2})|(a{3})

scala> "aa" match { case Regex(group1, group2) => (group1, group2); case _ => "no match" }
res1: java.io.Serializable = (aa,null)

scala> "aaa" match { case Regex(group1, group2) => (group1, group2); case _ => "no match" }
res2: java.io.Serializable = (null,aaa)

scala> "aaaa" match { case Regex(group1, group2) => (group1, group2); case _ => "no match" }
res3: java.io.Serializable = no match
```

benjben:
Superseded by #4076

PR #4053, opened 2019-04-26
Fix problems for IAB enrichment

benjben:
- Don't run enrichment if no user agent
- Don't run enrichment if IPv6
- Remove port IPv4


benjben:
Superseded by #4076

Issue #3951, opened 2019-01-21
Scala Common Enrich: support IPv6 addresses in IAB enrichment


chuwy:
We're seeing following bad row produced by IAB enrichment.

```
[...]

PR #4076, opened 2019-05-13
Release R114 Polonnaruwa

benjben:


benjben:
Good point for #3951 , I created #4078.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentManager.scala
```diff
@@ -279,14 +282,20 @@ object EnrichmentManager {
...
+    // and if the user agent is defined.
     val iabContext = registry.getIabEnrichment match {
       case Some(iab) =>
-        iab
-          .getIabContext(Option(event.useragent),
-                         Option(event.user_ipaddress),
-                         Option(event.derived_tstamp).map(EventEnrichments.fromTimestamp))
-          .map(_.some)
+        event.user_ipaddress match {
+          case IPv4Regex(ipv4) if !List(null, "", s"\0").contains(event.useragent) =>
```
BenFradet:
```diff
-- s"\0"
++ "\0"
```

also are those checks necessary given the regex match?

also, I'd rather we do `Option(event.user_ipaddress).map ...`

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/ConversionUtils.scala
```diff
@@ -42,7 +39,11 @@ import scalaz._
 import Scalaz._
 
 // Scala URI
-import com.netaporter.uri.Uri
+import io.lemonlabs.uri.Uri
+import io.lemonlabs.uri.Url
```
BenFradet: :art: we can combine those two imports

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentManager.scala
benjben: I'll update `"\0"`.

The regex is for the IP, while the checks are for the user agent.

benjben: I can't remove the `s` :
```
Octal escape literals are unsupported, use \u0000 instead
```

PR #4090, opened 2019-05-23
Release 115

benjben:


3-enrich/emr-etl-runner/lib/snowplow-emr-etl-runner/emr_job.rb
```diff
@@ -1109,7 +1109,7 @@ def wait_for
 
       # Prettified string containing failure details
       # for this job flow.
-      Contract String => String
+      Contract String, Elasticity::ClusterStatus, Elasticity::ClusterStepStatus => String
       def get_failure_details(jobflow_id, cluster_status, cluster_step_status_for_run)
```
BenFradet: since it's now pure you can easily write a unit test for it, shout if you need help doing that.

PR #4105, opened 2019-06-06
Scala Stream Collector: allow multiple cookie domains to be used (closes #3994)

### LINKED ISSUE #3994: Scala Stream Collector: allow multiple cookie domains to be used
The idea, if I got it right, would be to investigate the content of the `Origin` header and send back a `Set-Cookie` with the corresponding domain, e.g.:

```hocon
collector.cookie {
    enabled = true
    expiration =  365 days"
    name = cookie
    domains = {
      "snowplowanalytics.com" = ".snplow.com"
      "other-domain.com" = ".something-else.com"
      // we could have a fallback / catch all here
    }
}
```

cc @jbeemster

dilyand:


2-collectors/scala-stream-collector/examples/config.hocon.sample
```diff
@@ -55,10 +55,21 @@ collector {
...
+    # The domain is determined by matching the origin domains (from the Origin header of the request)
+    # to the list below. The first match is used. If no matches are found, the value will not be set
+    # and the collector's domain will be used.
+    # If the Origin header contains any subdomains, only the root domain names will be matched against
+    # the list below. That way, the cookie will be readable from all subdomains on the root domain.
+    domains = [
+        "{{cookieDomain1}}"
+        "{{cookieDomain2}}"
+        # ... more domains
+    ]
```
BenFradet: I think the original issue is around a mapping not just a list.

e.g. you get an incoming request with a `snplow.com` origin header, and you want to send back `snowplowanalytics.com` in your cookie.

Also, having given it a bit more thought, I think we need a `fallbackDomain` which is the mirroring of the current mechanism so that it is easier to transition.

dilyand: After chatting with @yalisassoon , he confirmed that a list of domains is enough. The only reason we would need a mapping would be to tie subdomains to the main domain. However, I have opted to instead allow a `*` wildcard to be used:

```
domains = [
  domain.com
  *.domain.com
]
```

The first one will match the main domain and the second one will match any subdomain on it.

@BenFradet could you expand on your other thought about the fallback? Do you mean that we need a default domain that will always be used, even if there are no matches from the Origin hosts?

BenFradet: >After chatting with @yalisassoon ...

then I think we need to modify the ticket name

>@BenFradet could you expand on your other thought about the fallback? Do you mean that we need a default domain that will always be used, even if there are no matches from the Origin hosts?

yes

dilyand: Done and done.

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
```diff
@@ -18,14 +18,12 @@ package collectors.scalastream
...
 
 import scala.collection.JavaConverters._
-
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.model.headers._
 import akka.http.scaladsl.model.headers.CacheDirectives._
 import org.apache.commons.codec.binary.Base64
 import org.slf4j.LoggerFactory
 import scalaz._
-
```
BenFradet: we tend to separate imports like this:

- java
- scala
- 3rd party
- 1st party

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
```diff
@@ -343,6 +342,74 @@ class CollectorService(
...
+            }
+          case _ => fallbackDomain
+        }
+      case None => fallbackDomain
+    }
+  }
+
+  /** Extracts the host names from a list of values in the request's Origin header. */
+  def extractHosts(origins: Seq[HttpOrigin]): Seq[String] = {
+    origins.map(origin => origin.host.host.address())
```
BenFradet: doesn't `origin.host.value` give us what we want?

BenFradet: :art: no need for curlies also

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
```diff
@@ -343,6 +342,74 @@ class CollectorService(
     if (pixelExpected) List(`Cache-Control`(`no-cache`, `no-store`, `must-revalidate`))
     else Nil
 
+  /**
+   * Determines the cookie domain to be used by inspecting the Origin header of the request
+   * and trying to find a match in the list of domains specified in the config file.
+   * The Origin header may include multiple domains. The first matching domain is used.
+   * If no match is found, the cookie domain is not set.
+   */
+  def cookieDomain(request: HttpRequest, domains: Option[List[String]], fallbackDomain: Option[String]): Option[String] = {
```
BenFradet: seems like you only need the headers and not the whole `HttpRequest`

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
```diff
@@ -343,6 +342,74 @@ class CollectorService(
     if (pixelExpected) List(`Cache-Control`(`no-cache`, `no-store`, `must-revalidate`))
     else Nil
 
+  /**
+   * Determines the cookie domain to be used by inspecting the Origin header of the request
+   * and trying to find a match in the list of domains specified in the config file.
+   * The Origin header may include multiple domains. The first matching domain is used.
+   * If no match is found, the cookie domain is not set.
+   */
```
BenFradet: It'd be great to have by-parameter scaladoc here I think

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
```diff
@@ -343,6 +342,74 @@ class CollectorService(
...
+  def toRegex(cookieDomainConfig: String): String = {
+    if (cookieDomainConfig.startsWith("*.")) {
+      val rawDomain = cookieDomainConfig.split("\\.").toList.drop(1).mkString("\\.")
+      ("*." + """(\.""" + rawDomain + ")").replace("*.", ".*")
+    } else {
+      val rawDomain = cookieDomainConfig.split("\\.").toList.mkString("\\.")
+      "(" + rawDomain + ")"
+    }
+  }
```
BenFradet: I'm not a fan of this post-processing, if the user is supposed to supply regexes, we shouldn't have to post-process them as it would lead to a lot of confusion

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
```diff
@@ -343,6 +342,74 @@ class CollectorService(
...
+  /**
+   * Determines the cookie domain to be used by inspecting the Origin header of the request
+   * and trying to find a match in the list of domains specified in the config file.
+   * The Origin header may include multiple domains. The first matching domain is used.
+   * If no match is found, the cookie domain is not set.
+   */
+  def cookieDomain(request: HttpRequest, domains: Option[List[String]], fallbackDomain: Option[String]): Option[String] = {
+    domains match {
+      case Some(domainList) =>
+        request.headers.find {
```
BenFradet: you can use `collect` here I think that'd simplify the code

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
```diff
@@ -343,6 +342,74 @@ class CollectorService(
...
+                  case _ => false
+                }
+                domainToUse.flatMap { d => Some(toDomain(d)) }
+              case None => fallbackDomain
+            }
+          case _ => fallbackDomain
+        }
+      case None => fallbackDomain
+    }
+  }
```
BenFradet: I think this could turn into a big for comp, no? that'd simplify the code quite a bit I think

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
dilyand: > doesn't `origin.host.value` give us what we want?

No, it returns the full host, including the port.

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
dilyand: I replaced the regex matching with simpler substring matching, and removed the `toRegex` and `toDomain` post-processing.

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
dilyand: Done.

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
dilyand: Done.

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
dilyand: Done.

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
dilyand: Done.

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
dilyand: Done.

dilyand:
# Integration tests

## Cookie domains -- not configured; fallback domain -- not configured

No `domain` directive should be set in the `Set-Cookie` header.

```
curl -i -X POST -H "Origin: http://origin.com" http://localhost:12345/com.snowplowanalytics.snowplow/tp2 -d '{"aid": "test"}'

HTTP/1.1 200 OK
Set-Cookie: sp=d7dec9fe-2c41-4ec9-920f-0481c0aeca49; Expires=Fri, 12 Jun 2020 17:36:57 GMT; Path=/
```

## Cookie domains -- not configured; fallback domain -- configured

```
fallbackDomain = "fallback.com"
```

The fallback domain should be used in the `domain` directive of the `Set-Cookie` header.

```
curl -i -X POST -H "Origin: http://origin.com" http://localhost:12345/com.snowplowanalytics.snowplow/tp2 -d '{"aid": "test"}'

HTTP/1.1 200 OK
Set-Cookie: sp=624cb9a3-e11d-437a-8177-53445c759b82; Expires=Fri, 12 Jun 2020 17:41:03 GMT; Domain=fallback.com; Path=/
```

## Cookie domains -- configured; fallback domain -- not configured

```
domains = [
  "snplow.com"
  "snowplow.com"
]
```

### The request's `Origin` header has a single domain that has a match in the list.

#### The origin domain is an exact match.

The first matching domain should be used in the `domain` directive of the `Set-Cookie` header.

```
curl -i -X POST -H "Origin: http://snowplow.com" http://localhost:12345/com.snowplowanalytics.snowplow/tp2 -d '{"aid": "test"}'

HTTP/1.1 200 OK
Set-Cookie: sp=18a3edcf-bba4-4750-b2f7-84005945f6af; Expires=Fri, 12 Jun 2020 17:49:29 GMT; Domain=snowplow.com; Path=/
```

#### The origin domain is a subdomain of a configured cookie domain.

The first matching configured cookie domain should be used in the `domain` directive of the `Set-Cookie` header.

```
curl -i -X POST -H "Origin: http://www.snowplow.com" http://localhost:12345/com.snowplowanalytics.snowplow/tp2 -d '{"aid": "test"}'

HTTP/1.1 200 OK
Set-Cookie: sp=bce00b03-4a20-4102-a3d7-621f7f4ec217; Expires=Fri, 12 Jun 2020 17:52:31 GMT; Domain=snowplow.com; Path=/
```

### The request's `Origin` header has multiple domains but only one has a match in the list.

The first matching configured cookie domain should be used in the `domain` directive of the `Set-Cookie` header.

```
curl -i -X POST -H "Origin: http://www.snowplow.com http://notsnowplow.com:8080" http://localhost:12345/com.snowplowanalytics.snowplow/tp2 -d '{"aid": "test"}'

HTTP/1.1 200 OK
Set-Cookie: sp=c84d3779-4470-4e9e-b28e-80e055df4990; Expires=Fri, 12 Jun 2020 17:58:20 GMT; Domain=snowplow.com; Path=/

curl -i -X POST -H "Origin: http://notsnowplow.com http://www.snowplow.com:8080" http://localhost:12345/com.snowplowanalytics.snowplow/tp2 -d '{"aid": "test"}'

HTTP/1.1 200 OK
Set-Cookie: sp=507ea94e-1cce-4d8d-ae47-26b15d663caa; Expires=Fri, 12 Jun 2020 17:59:14 GMT; Domain=snowplow.com; Path=/
```

### The request's `Origin` header has multiple domains that have matches in the list.

The first matching configured cookie domain should be used in the `domain` directive of the `Set-Cookie` header.

```
curl -i -X POST -H "Origin: http://notsnplow.com http://www.snplow.com http://snowplow.com:8080" http://localhost:12345/com.snowplowanalytics.snowplow/tp2 -d '{"aid": "test"}'

HTTP/1.1 200 OK
Set-Cookie: sp=a1077f8c-a617-4711-9d2e-2e7c704c2396; Expires=Fri, 12 Jun 2020 18:01:45 GMT; Domain=snplow.com; Path=/
```

### None of the domains in the request's `Origin` header has a match in the list.

No `domain` directive should be set in the `Set-Cookie` header.

```
curl -i -X POST -H "Origin: http://notsnplow.com http://www.notsnplow.com http://notsnowplow.com:8080" http://localhost:12345/com.snowplowanalytics.snowplow/tp2 -d '{"aid": "test"}'

HTTP/1.1 200 OK
Set-Cookie: sp=b4df17b6-830a-476c-a13c-a40dd32c18c2; Expires=Fri, 12 Jun 2020 18:03:34 GMT; Path=/
```

## Cookie domains -- configured; fallback domain -- configured

```
domains = [
  "snplow.co.uk"
  "snowplow.co.uk"
]
fallbackDomain = "fallback.com"
```

### None of the domains in the request's `Origin` header has a match in the list.

The fallback domain should be used in the `domain` directive of the `Set-Cookie` header.

```
curl -i -X POST -H "Origin: http://notsnplow.co.uk http://www.notsnplow.co.uk http://notsnowplow.co.uk:8080" http://localhost:12345/com.snowplowanalytics.snowplow/tp2 -d '{"aid": "test"}'

HTTP/1.1 200 OK
Set-Cookie: sp=3c7a1ece-c8b2-4e6d-95d1-01866336ba4a; Expires=Fri, 12 Jun 2020 18:11:45 GMT; Domain=fallback.com; Path=/
```

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
```diff
@@ -343,6 +347,47 @@ class CollectorService(
...
+    (for {
+      domainList <- domains
+      origins <- headers.collectFirst { case header: `Origin` => header.origins }
+      originHosts = extractHosts(origins)
+      matchingDomains = for {
+        host <- originHosts
+        matchingDomains <- domainList.collectFirst { case domain if validMatch(host, domain) => domain }
+      } yield matchingDomains
+      domainToUse <- matchingDomains.headOption
+    } yield domainToUse) match {
```
BenFradet: you can do an `orElse(fallbackDomain)` here instead of the pattern matching but otherwise it looks great! :+1:

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
```diff
@@ -343,6 +347,47 @@ class CollectorService(
...
+   */
+  def cookieDomain(headers: Seq[HttpHeader], domains: Option[List[String]], fallbackDomain: Option[String]): Option[String] =
+    (for {
+      domainList <- domains
+      origins <- headers.collectFirst { case header: `Origin` => header.origins }
+      originHosts = extractHosts(origins)
+      matchingDomains = for {
+        host <- originHosts
+        matchingDomains <- domainList.collectFirst { case domain if validMatch(host, domain) => domain }
+      } yield matchingDomains
```
BenFradet: do you need the nested for-comp?

Maybe you can simply do:

```scala
originHosts = extractHosts(origins)
domainToUse <- domainList.filter(domain => originHosts.exists(validMatch(_, domain)).headOption
```

wdyt?

2-collectors/scala-stream-collector/core/src/test/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorServiceSpec.scala
```diff
@@ -401,5 +401,75 @@ class CollectorServiceSpec extends Specification {
...
+          val cookieConfig = CookieConfig(true, "name", 5.seconds, None, None)
+          service.cookieDomain(request.headers, cookieConfig.domains, cookieConfig.fallbackDomain) shouldEqual None
+        }
+        "if a list of domains is supplied in the config but the Origin request header is empty and there is no fallback domain" in {
+          val request = HttpRequest()
+          val cookieConfig = CookieConfig(true, "name", 5.seconds, Some(List("domain.com")), None)
+          service.cookieDomain(request.headers, cookieConfig.domains, cookieConfig.fallbackDomain) shouldEqual None
+        }
+        "if none of the domains in the request's Origin header has a match in the list of domains supplied with the config and there is no fallback domain" in {
+          val origins = scala.collection.immutable.Seq(HttpOrigin("http", Host("origin.com")), HttpOrigin("http", Host("otherorigin.com", 8080)))
```
BenFradet: :art: you can go with just `Seq` or `List` in tests that will make the code more readable I think

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
dilyand: Brilliant, looks much better now.

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
```diff
@@ -343,6 +345,40 @@ class CollectorService(
...
+   * Ensures a match is valid.
+   * We only want matches where:
+   * a.) the Origin host is exactly equal to the cookie domain from the config
+   * b.) the Origin host is a subdomain of the cookie domain from the config.
+   * But we want to avoid cases where the cookie domain from the config is randomly
+   * a substring of the Origin host, without any connection between them.
+   */
+  def validMatch(host: String, domain: String): Boolean =
+    if (host == domain || host.endsWith("." + domain)) true
+    else false
```
BenFradet: you can just do

```scala
def validMatch(host: String, domain: String): Boolean =
  host == domain || host.endsWith("." + domain)
```

dilyand: Done.

PR #4111, opened 2019-06-24
Scala Stream Collector: provide a way to specify custom path mappings (closes #4087)

### LINKED ISSUE #4087: Scala Stream Collector: provide a way to specify custom path mappings
e.g.

```
com.acme/track -> com.snowplowanalytics.snowplow/tp2
com.acme/redirect -> r/tp2
com.acme/iglu -> com.snowplowanalytics.iglu/v1
```

dilyand:


2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala
```diff
@@ -127,6 +127,7 @@ package model {
   final case class CollectorConfig(
     interface: String,
     port: Int,
+    paths: Map[String, String] = Map.empty[String, String],
```
chuwy: Default parameters generally considered a bad practice in Scala https://blog.ssanj.net/posts/2019-05-01-why-are-default-parameter-values-considered-bad-in-scala.html. Especially if they're not in the end.

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
```diff
@@ -78,6 +79,16 @@ class CollectorService(
   override val cookieName = config.cookieName
   override val doNotTrackCookie = config.doNotTrackHttpCookie
 
+  /**
+   * Determines the path to be used in the response,
+   * based on whether a mapping can be found in the config for the original request path.
+   */
+  override def determinePath(vendor: String, version: String): String = {
+    val original = s"$vendor/$version"
+    val mappingOrOriginal = config.paths.getOrElse(original, original)
```
chuwy: Is there a reason why we don't have a slash in the beginning in the mapping?

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorRoute.scala
```diff
@@ -43,7 +42,7 @@ trait CollectorRoute {
           extractors { (host, ip, request) =>
             // get the adapter vendor and version from the path
             path(Segment / Segment) { (vendor, version) =>
-              val path = s"/$vendor/$version"
+              val path = collectorService.determinePath(vendor, version)
```
chuwy: So, does it mean we can rewrite only two-segmented URIs with two-segmented URIs? `/com.acme/foo/bar` won't be supported neither as a source nor destinatioin?

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
dilyand: Do you mean in the config? I think it's more user-friendly and less error-prone without it. Compare:

```
paths {
  "/com.acme/track" -> "/com.snowplowanalytics.snowplow/tp2
}
```

vs

```
paths {
  "com.acme/track" -> "com.snowplowanalytics.snowplow/tp2
}
```

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorRoute.scala
dilyand: Yes. This is the current behaviour as well.

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala
dilyand: Changed it to `Option[Map[String, String]]`.

dilyand:
Also, bumped version to `0.16.0-rc2`, as there already are published artefacts for `0.16.0-rc1`.

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala
chuwy: But `None` is not different from empty `Map`?

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
chuwy: That's arguable. For me `/com.acme/track` looks explicit in the sense that is a root-level URL. Though it is a nit, I don't have strong opinion about it.

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala
dilyand: Sure, but then:

```
val mappings = config.paths.getOrElse(Map.empty[String, String])
```

chuwy: Exactly. And we're ending up with empty map by default anyways, unless I'm missing something very obvious.

dilyand: Maybe I am...

Your note was about default parameters, especially when not last in the list. We are no longer using them. The `paths` parameter is an optional map. You cannot look for a value by key in a `None`, so if the map is not provided, we use an empty map instead.

Do you still have any concerns?

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala
```diff
@@ -127,6 +127,7 @@ package model {
   final case class CollectorConfig(
     interface: String,
     port: Int,
+    paths: Option[Map[String, String]],
```
BenFradet: I think we'd want to make it mandatory as options don't play well with configuration templating

2-collectors/scala-stream-collector/examples/config.hocon.sample
```diff
@@ -24,6 +24,22 @@ collector {
...
+  # - com.snowplowanalytics.snowplow/tp2 for Tracker Protocol 2
+  # - r/tp2 for redirects
+  # - com.snowplowanalytics.iglu/v1 for the Iglu Webhook
+  # Any path that matches the 'vendor/version' protocol will result in a cookie response, for use by custom webhooks
+  # downstream of the collector.
+  # But you can optionally also map any valid (i.e. two-segment) path to one of the three defaults.
+  # Your custom path must be the key and the value must be one of the corresponding default paths.
+  # Uncomment to use. Defaults to empty map.
+  #paths {
+  #  "com.acme/track" = "com.snowplowanalytics.snowplow/tp2"
```
BenFradet: I tend to agree with Anton, we're better off with fully qualified paths

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorRoute.scala
BenFradet: did you look into [Segments](https://doc.akka.io/docs/akka-http/current/routing-dsl/path-matchers.html) in the end to allow for arbitrary incoming paths?

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala
dilyand: What is users don't want to map anything? Would we require them to pass in the config:

```
paths {
  "r/tp2" = "r/tp2"
}
```

?

dilyand:
I did look into `Segments` and even implemented an arbitrary-length path but then decided to scrap it because it is unclear how we should handle unmapped paths. I summarised the issue here: https://github.com/snowplow/snowplow/issues/4087#issuecomment-505056435 .

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala
BenFradet: if they don't want to map anything they would specify `paths { }`

dilyand: Ah, and then that would be parsed as `Map.empty`?

2-collectors/scala-stream-collector/examples/config.hocon.sample
dilyand: I've changed it to the way you guys advise.

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala
dilyand: Done.

BenFradet:
unmapped paths should be left alone no?

dilyand:
By 'left alone' do you mean passed on? That is not the current behaviour. The current behaviour is:

```
com.acme => 404
com.acme/random => 200, path=com.acme/random
com.acme/random/other => 404
```

I have preserved this behaviour in `0.16.0-rc2`. We apply the same rules to all values, regardless whether they are mapped or not. The only difference is that if there is a mapping, we swap it in before doing further processing.

If the behaviour needs to be changed, i.e. if we need different rules for mapped and non-mapped values, then there are a couple of questions to answer:
- Which type of values should inherit the current rules -- mapped or non-mapped?
- What should be the rules for the other type of values?

Since nobody was able to answer these questions, I parked them for further discussion down the line.

Does that make sense @BenFradet?

MrMoronIV:
Say I change the default path to something else, how do I make stream enrich play along?

It seems they show it towards the bad topic with the error: 
```
Payload with vendor com.info.nl and version info2 not supported by this version of Scala Common Enrich
```

This is my complete question on the discourse: https://discourse.snowplowanalytics.com/t/collector-path-change-stream-enrich-doesnt-like-it/3288

markharding:
Is this change available on https://bintray.com/snowplow/registry/snowplow%3Ascala-stream-collector-kinesis ?

benjben:
Hi @markharding , latest collector version is 1.0.1, you can find it on [Docker Hub](https://hub.docker.com/layers/snowplow/scala-stream-collector-kinesis/1.0.1/images/sha256-3839e0be14e6095f19cef08e034012789e548243ef225a9809d74d65fdab06b1?context=explore).

We're not publishing images to Bintray any more.

Issue #3487, opened 2017-10-30
EmrEtlRunner: add support to the spot market for core instances


bernardosrulzon:
This one is quite straightforward: supporting the spot market for core instances can reduce EMR costs significantly (~60% in our case), which is especially important now that the long de-duplication and RDB Loader steps were moved to EMR. The open question is whether data would be lost in case of an instance termination.

PS: Task nodes are not effective as rbolkey [pointed out on the forums](https://discourse.snowplowanalytics.com/t/learnings-from-using-the-new-spark-emr-jobs/1260).

alexanderdean:
Thanks @bernardosrulzon . We will keep this ticket open to track the requirement, but it's very unlikely to be implemented before the move to Dataflow Runner (at which point we will migrate this ticket).

bernardosrulzon:
That makes a lot of sense @alexanderdean.

I guess the key issue is whether Snowplow is ready for spot instances or there's additional development involved in making it resistant to instance termination/rebooting - which would be especially troublesome on the Shred Step as data would already be inserted into DynamoDB.

Any thoughts on that? /cc @BenFradet

BenFradet:
Unfortunately cross-batch deduplication through DynamoDB is not idempotent. As a result, we'd need to make the two exclusive.

This kind of makes sense since the two are going in different directions, cross-batch deduplication is a sizable added cost whereas when using spot instances you're trying to save money.

chuwy:
As we re-discovered recently - cross-batch deduplication is idempotent. When you're shredding even that is already in manifest **and has same etl timestamp** it will not be marked as duplicate. It means that we can safely re-deduplicate events on spot instances without loosing them.

bernardosrulzon:
That's the point @chuwy - when a job fails in Shred and we re-process data using EmrEtlRunner, a new job is created with an updated `etl_tstamp`, which causes the previous events to be discarded as duplicates

PR #4123, opened 2019-08-02
Allow EmrEtlRunner to use spot-priced core instances

miike:
This adds support for spot pricing core instances (and addresses #3487).

Let me know if I should also add a companion PR to bump the Dataflow runner cluster config schema to 1-0-1 to include the core bid field (I've left it as is for the tests).


3-enrich/emr-etl-runner/config/config.yml.sample
```diff
@@ -45,6 +45,7 @@ aws:
       master_instance_type: m1.medium
       core_instance_count: 2
       core_instance_type: m1.medium
+      core_instance_bid: 0.015
```
chuwy: Should we add same comment as in `stream_config.yml.sample` here?

ian-axelrod: I'm curious (forgive me if I'm wrong for making a comment here), but is there a reason you are requiring spot instance bids? I would expect this to be annoying to maintain. Does EMR not support a 'find lowest bid' approach, similar to several other AWS services (e.g., autoscaling groups)? Would it not be feasible to implement one? Cheers!

benjben: Hi @ian-axelrod , thank you for making a comment. As you can see [here](https://github.com/snowplow/snowplow/pull/4123/files/4554517a3d5da24470335b3286e8f120a1bd71ea#diff-123bd97fb90bffdf6bcd93c5eb634da7), this change only gives the possibility to use spot instance bids, it doesn't force it. Some users of Snowplow pipeline have a need for it.

benjben:
Got merged in 922aaa73852bf807f334394c80298b324972b383

PR #4204, opened 2019-09-26
Scala Stream Collector: Adds support fo TLS-encryption (fixes #4085)

### LINKED ISSUE #4085: Scala Stream Collector: add support for TLS port binding and certificate
At the moment the SSC only binds over an unsecured port which means we cannot achieve end-to-end encryption when tracking data.  It would be great to add an optional TLS port for the collector so that we could achieve this ability.

peel:
The change introduces an additional parameter set that allows for enabling secure TLS encryption over HTTPS. All major features exposed by `lightbend/ssl-config` (upon which akka-http TLS relies)
are implemented. To enable, define:
```
collector.ssl {
  enable = true # defaults to false
  redirect = true # defaults to false - whether to upgrade http to https
  port = 443 # defaults to 443 
}
```

The change requires update for akka-http (`10.0.x` -> `10.1.x`) as 10.0.x series fails to provide TLS handshake.

TLS setup is a mutable java API and it is clearly visible.


peel:
I added a simple test case under `./integration-tests/ssc-ssl-test.sh` that drags around two separate configs and the certs. It will sequentially start the service without and with SSL. Let me know if there's a better place to put these scripts.

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala
```diff
@@ -130,6 +136,7 @@ package model {
   final case class CollectorConfig(
     interface: String,
     port: Int,
+    ssl: Boolean,
```
benjben: Would it be possible to make this parameter optional in the configuration file and to give it a default value "false" if absent? It would make TechOps' work easier for R117, in that it wouldn't require to update the configuration files when people don't want to activate SSL.

.travis.yml
```diff
@@ -12,6 +12,7 @@ script:
   - python integration-tests/api-lookup-test.py &
   - cd 3-enrich/scala-common-enrich && sbt publishLocal && cd ../..
   - cd $TEST_DIR && sbt test && cd ../..
+  - test "${TEST_DIR}" == "2-collectors/scala-stream-collector/" && cd $TRAVIS_BUILD_DIR/integration-tests && bash -c ./ssc-ssl-test.sh && cd - || true
```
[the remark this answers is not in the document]
peel: Yes it will, good catch.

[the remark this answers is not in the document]
benjben: Perfect!

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/Collector.scala
```diff
@@ -87,12 +90,40 @@ trait Collector {
...
-          s"${collectorConf.interface}:${collectorConf.port}", ex.getMessage)
+    lazy val redirectRoutes =
+      scheme("http") {
+        extract(_.request.uri) { uri =>
+          redirect(
+            uri.copy(scheme = "https").withPort(collectorConf.portHttps),
+            StatusCodes.MovedPermanently
+          )
+        }
       }
```
benjben: Does it mean that when SSL is activated, an HTTP request will automatically be redirected to the HTTPs endpoint?

peel:
I added a http -> https redirect to ease on ops. Hope it’s still fine.

jbeemster:
@peel is it possible to make that an optional switch to do the redirect?

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/Collector.scala
peel: Exactly that. It will match only http,  switch scheme to https and update the port and eventually redirect to the new url.

peel:
@jbeemster a mode that exposes the same routes both on https and http? Sure!

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/Collector.scala
benjben: Awesome :+1:

jbeemster:
@peel that looks great!  Cannot wait to get my hands on it!

benjben:
LGTM!
This is great stuff!

integration-tests/no-ssl.hocon
```diff
@@ -0,0 +1,332 @@
...
+    # If metrics are enabled then all requests will be logged as prometheus metrics
+    # and '/metrics' endpoint will return the report about the requests
+    enabled = false
+    enabled = ${?COLLECTOR_PROMETHEUS_METRICS_ENABLED}
+    # Custom buckets for http_request_duration_seconds_bucket duration metric
+    #durationBucketsInSeconds = [0.1, 3, 10]
+    #durationbucketsInSeconds = ${?COLLECTOR_PROMETHEUS_METRICS_DURATION_BUCKETS_IN_SECONDS}
+  }
+
+  # Configuration of prometheus http metrics
```
dilyand: The prometheus metrics are duplicated in the original hocon example and this has been carried over here. I'll open a separate issue to clean up the hocon, but maybe we should get rid of this duplication for the test as well.

integration-tests/with-ssl.hocon
```diff
@@ -0,0 +1,358 @@
...
+    # If metrics are enabled then all requests will be logged as prometheus metrics
+    # and '/metrics' endpoint will return the report about the requests
+    enabled = false
+    enabled = ${?COLLECTOR_PROMETHEUS_METRICS_ENABLED}
+    # Custom buckets for http_request_duration_seconds_bucket duration metric
+    #durationBucketsInSeconds = [0.1, 3, 10]
+    #durationbucketsInSeconds = ${?COLLECTOR_PROMETHEUS_METRICS_DURATION_BUCKETS_IN_SECONDS}
+  }
+
+  # Configuration of prometheus http metrics
```
dilyand: Duplication.

PR #4212, opened 2019-10-07
Scala Stream Collector: allow disabling redirects (fixes #4211)

### LINKED ISSUE #4211: Scala Stream Collector: allow users to disable the default redirect endpoint
The Javascript tracker allows for click tracking through open redirects. There are concerns that this functionality might be abused by malicious third parties. Even if SSC users are not making use of the feature in the JS tracker, they are still potentially vulnerable as the collector would accept requests to the default redirect endpoint.

peel:
Add a new flag `collector.allowRedirects` that when set to false will prohibit (`410 Gone`) any redirects using `r/` prefix url.

Not adding `Bump to version`. This probably should come with the same version as the #4204.

### DIFF
--- a/2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorRoute.scala
+++ b/2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorRoute.scala
@@ -33,7 +33,15 @@ trait CollectorRoute {
   def extractContentType: Directive1[ContentType] =
     extractRequestContext.map(_.request.entity.contentType)
 
-  def collectorRoute: Route =
+  def collectorRoute =
+    if (collectorService.allowRedirects) routes else rejectRedirect ~ routes
+
+  def rejectRedirect: Route =
+    path("r" / Segment) { _ =>
+      complete(StatusCodes.NotFound -> "redirects disabled")
+    }
+
+  def routes: Route =
     doNotTrack(collectorService.doNotTrackCookie) { dnt =>
       cookieIfWanted(collectorService.cookieName) { reqCookie =>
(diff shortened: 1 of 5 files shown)


2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorRoute.scala
```diff
@@ -33,7 +33,15 @@ trait CollectorRoute {
   def extractContentType: Directive1[ContentType] =
     extractRequestContext.map(_.request.entity.contentType)
 
-  def collectorRoute: Route =
+  def collectorRoute =
+    if (collectorService.allowRedirects) routes else rejectRedirect ~ routes
+
+  def rejectRedirect: Route =
+    path("r" / Segment) { _ =>
+      complete(StatusCodes.Gone -> "redirects disabled")
```
lukeindykiewicz: Why `410` actually? 
> The HTTP specification intends this status code to be used for "limited-time, promotional services". APIs should not feel compelled to indicate resources that have been deleted with this status code.

Is this true for this redirects?
This is from https://developer.mozilla.org/en-US/docs/Web/HTTP/Status

Maybe simple `404`, wdyt?

dilyand:
Just a nit and can be fixed later as well: the commit name should be the same as the issue.

2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorRoute.scala
peel: Interesting, haven't heard about "limited-time, promotional offers" part. `404` makes sense.

PR #4223, opened 2019-10-16
Scala Common Enrich: add anonymization for IPv6 (fixes #4222)

### LINKED ISSUE #4222: Scala Common Enrich: add anonymization for IPv6

lukeindykiewicz:


3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
```diff
@@ -52,23 +55,30 @@ object AnonIpEnrichment extends ParseableEnrichment {
...
+    def extractParam(name: String) =
+      ScalazJson4sUtils.extract[Int](config, "parameters", name)
+
+    isParseable(config, schemaKey).flatMap(_ => {
       (for {
-        param  <- ScalazJson4sUtils.extract[Int](config, "parameters", "anonOctets")
-        octets <- AnonOctets.fromInt(param)
-        enrich = AnonIpEnrichment(octets)
+        paramIPv4Octet   <- extractParam("anonOctets")
+        paramIPv6Segment <- extractParam("anonSegments").orElse(paramIPv4Octet.success)
```
lukeindykiewicz: This `default` value for number of anonymized segments is provided to not break backwards compatibility with "patch" release, which R117 was meant to be.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
```diff
@@ -118,23 +167,60 @@ case class AnonIpEnrichment(
...
+
+    if (ip.startsWith(IPv4MappedAddressPrefix))
+      IPv4MappedAddressPrefix + masked(ip.replace(IPv4MappedAddressPrefix, ""))
+    else masked(ip)
+  }
+
+  private def anonymizeIpV6(ip: String): String =
+    InetAddress
+      .getByName(ip)
+      .getHostAddress
```
lukeindykiewicz: `InetAddress.getByName(ip).getHostAddress` when provided with `ip` (not `host name`) does not do any dns lookup calls. It's used here to not have to deal with tricky IPv6 format, which allows omitting certain parts. `hetHostName` returns always 8 segments, so it makes it easy to anonymize.

3-enrich/scala-common-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/registry/AnonIpEnrichmentSpec.scala
```diff
@@ -23,20 +23,35 @@ import org.specs2.matcher.DataTables
...
+      "ipv6, anonymize 3"    !! "2605:2700:0:3:0:0:4713:93e3"   ! AnonIPv4Octets(3)       ! AnonIPv6Segments(3)     ! "2605:2700:0:3:0:x:x:x" |
+      "ipv6, anonymize 4"    !! "2605:2700:0:3::4713:93e3"      ! AnonIPv4Octets(4)       ! AnonIPv6Segments(4)     ! "2605:2700:0:3:x:x:x:x" |
+      "ipv6, anonymize 5"    !! "2605:2700:0:3::4713:93e3"      ! AnonIPv4Octets(4)       ! AnonIPv6Segments(5)     ! "2605:2700:0:x:x:x:x:x" |
+      "ipv6, anonymize 6"    !! "2605:2700:0:3::4713:93e3"      ! AnonIPv4Octets(4)       ! AnonIPv6Segments(6)     ! "2605:2700:x:x:x:x:x:x" |
+      "ipv6, anonymize 7"    !! "2605:2700:0:3::4713:93e3"      ! AnonIPv4Octets(4)       ! AnonIPv6Segments(7)     ! "2605:x:x:x:x:x:x:x" |
+      "ipv6, anonymize 8"    !! "2605:2700:0:3::4713:93e3"      ! AnonIPv4Octets(4)       ! AnonIPv6Segments(8)     ! "x:x:x:x:x:x:x:x" |
+      "ipv6, anonymize 8"    !! "2605:2700::4713:93e3"          ! AnonIPv4Octets(4)       ! AnonIPv6Segments(8)     ! "x:x:x:x:x:x:x:x" |
+      "ipv6, anonymize 8"    !! "2700::4713"                    ! AnonIPv4Octets(4)       ! AnonIPv6Segments(8)     ! "x:x:x:x:x:x:x:x" |
```
lukeindykiewicz: I've made an arbitrary decision to return IPv6 in not shortened version. It simplifies development and makes it visible that some anonymization took place.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
```diff
@@ -77,22 +87,61 @@ object AnonOctets extends Enumeration {
...
    *
-   * @param anonOctets A String holding
+   * @param anonIPv4Octets A String holding
    *        the number of IP address
    *        octets to anonymize
-   * @return a Validation-boxed AnonOctets
+   * @return a Validation-boxed AnonIPv4Octets
    */
-  def fromInt(anonOctets: Int): ValidatedMessage[AnonOctets] =
+  def fromInt(anonIPv4Octets: Int): ValidatedMessage[AnonIPv4Octets] =
```
peel: Technically the type is `ValidatedMessage[_] \/ Throwable`. Maybe it's better to have it fully `ValidatedMessage` and return a generic failure.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
```diff
@@ -77,22 +87,61 @@ object AnonOctets extends Enumeration {
...
+   * Update the Validation Error if the
+   * conversion isn't possible.
+   *
+   * @param anonSegments A String holding
+   *        the number of IPv6 address
+   *        segments to anonymize
+   * @return a Validation-boxed AnonIPv6Segments
+   */
+  def fromInt(anonIPv6Segments: Int): ValidatedMessage[AnonIPv6Segments] =
+    try {
```
peel: I'd stay away from `try`

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
lukeindykiewicz: Totally agree, but similar code is in all enrichments, so I wanted to keep it consistent.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
lukeindykiewicz: Totally agree, but similar code is in all enrichments, so I wanted to keep it consistent.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
```diff
@@ -102,14 +151,14 @@ object AnonOctets extends Enumeration {
  * @param octets The number of octets to anonymize
```
dilyand: The Scaladoc needs to be updated to reflect the changes.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
```diff
@@ -102,14 +151,14 @@ object AnonOctets extends Enumeration {
...
+  ipv4Octets: AnonIPv4Octets.AnonIPv4Octets,
+  ipv6Segments: AnonIPv6Segments.AnonIPv6Segments
 ) extends Enrichment {
 
   /**
    * Anonymize the supplied IP address.
    *
-   * octets is the number of octets
-   * in the IP address to anonymize, starting
+   * octets is the number of octets in the IP address to anonymize, starting
```
dilyand: The name of this param has changed, this must be reflected in the Scaladoc.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
```diff
@@ -118,23 +167,60 @@ case class AnonIpEnrichment(
    * anonymizeIp("94.15.223.151", Three)
    * => "94.x.x.x"
    *
+   *
+   * For IPv6 either the form defined in RFC 2732
```
dilyand: Should we add examples like for IPv4? Maybe we can also get rid of the first example for IPv4, to make the Scaladoc more concise.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
```diff
@@ -118,23 +167,60 @@ case class AnonIpEnrichment(
...
+        .map {
+          case _: Inet4Address => anonymizeIpV4(ip)
+          case _: Inet6Address => anonymizeIpV6(ip)
+        }
+        .getOrElse(tryAnonymizingInvalidIp(ip))
+    }.orNull
+
+  private def anonymizeIpV4(ip: String): String = {
+    def masked(ipv4: String) = {
+      val splitted = ipv4.split("\\.")
```
dilyand:
```suggestion
      val split = ipv4.split("\\.")
```

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
```diff
@@ -118,23 +167,60 @@ case class AnonIpEnrichment(
...
+      splitted
```
dilyand:
```suggestion
      split
```

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
```diff
@@ -118,23 +167,60 @@ case class AnonIpEnrichment(
...
+        .take(AnonIPv4Octets.All.id - ipv4Octets.id)
+        .toList
+        .padTo(splitted.size, MaskChar)
```
dilyand:
```suggestion
        .padTo(split.size, MaskChar)
```

3-enrich/scala-common-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/registry/AnonIpEnrichmentSpec.scala
dilyand: 👍🏻

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
```diff
@@ -118,23 +167,60 @@ case class AnonIpEnrichment(
...
+      .split(":")
+      .take(AnonIPv6Segments.All.id - ipv6Segments.id)
+      .toList
+      .padTo(AnonIPv6Segments.All.id, MaskChar)
+      .mkString(":")
+
+  /**
+   * Mainly to not brake code that already exists, i.e. broken IP like this: "777.2.23"
+   * */
+  private def tryAnonymizingInvalidIp(ip: String): String =
```
dilyand: What if we have an invalid IPv6?

3-enrich/scala-common-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/registry/EnrichmentConfigsSpec.scala
```diff
@@ -63,7 +63,26 @@ class EnrichmentConfigsSpec extends Specification with ValidationMatchers with D
       val schemaKey = SchemaKey("com.snowplowanalytics.snowplow", "anon_ip", "jsonschema", "1-0-0")
```
dilyand: We'll need a new schema for this in `iglu-central`.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
lukeindykiewicz: It's not possible to have invalid IPv6 I think. The basic cases (i.e. null, just string) are handled by IPv4 (you can not know if this was meat to be IPv6). When you have a missing segment in IPv6 the `InetAddress.getByName` deals with it - that's why I'm using it. 
Ok, I just came up with one more broken case, like 123z4::2653 (string that contains not hex number char in it). I'll add test and some fix for it.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
lukeindykiewicz: the `val` contains already splitted ipv4, that's why it's splitted, not split.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
benjben: I've seen somewhere on Zendesk that some customers receive IPv6s with `.` instead of `:`. Would such IPs succeed ?

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
```diff
@@ -52,23 +55,30 @@ object AnonIpEnrichment extends ParseableEnrichment {
...
+  }
 
 }
 
 /**
- * How many octets to anonymize?
+ * How many octets (ipv4) to anonymize?
  */
-object AnonOctets extends Enumeration {
+object AnonIPv4Octets extends Enumeration {
```
chuwy: IMO its a common convention to have only class/trait `Foo` and object `Foo` in a module called `Foo.scala`. I'd move it inside `AnonIpEnrichment`. Also I didn't expect to see Scala `Enumeration` in SCE, we usually encode enums as sealed traits with objects.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
```diff
@@ -77,22 +87,61 @@ object AnonOctets extends Enumeration {
...
    */
-  def fromInt(anonOctets: Int): ValidatedMessage[AnonOctets] =
+  def fromInt(anonIPv4Octets: Int): ValidatedMessage[AnonIPv4Octets] =
     try {
-      AnonOctets(anonOctets).success
+      AnonIPv4Octets(anonIPv4Octets).success
     } catch {
-      case nse: NoSuchElementException => "IP address octets to anonymize must be 1, 2, 3 or 4".toProcessingMessage.fail
+      case nse: NoSuchElementException =>
+        "IPv4 address octets to anonymize must be 1, 2, 3 or 4".toProcessingMessage.fail
```
chuwy: I'd mention what was given:

```scala
s"IPv4 address octets to anonymize must be 1, 2, 3 or 4. $anonIPv4Octets is given"
```

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
```diff
@@ -77,22 +87,61 @@ object AnonOctets extends Enumeration {
...
+   * @return a Validation-boxed AnonIPv4Octets
    */
-  def fromInt(anonOctets: Int): ValidatedMessage[AnonOctets] =
+  def fromInt(anonIPv4Octets: Int): ValidatedMessage[AnonIPv4Octets] =
     try {
-      AnonOctets(anonOctets).success
+      AnonIPv4Octets(anonIPv4Octets).success
     } catch {
-      case nse: NoSuchElementException => "IP address octets to anonymize must be 1, 2, 3 or 4".toProcessingMessage.fail
+      case nse: NoSuchElementException =>
```
chuwy:
```scala
case _: NoSuchElementException =>
```

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
```diff
@@ -77,22 +87,61 @@ object AnonOctets extends Enumeration {
...
+        "IPv4 address octets to anonymize must be 1, 2, 3 or 4".toProcessingMessage.fail
+    }
+}
+
+/**
+ * How many segments (ipv6) to anonymize?
+ */
+object AnonIPv6Segments extends Enumeration {
```
chuwy: Same about module and enumeration.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
```diff
@@ -118,23 +167,60 @@ case class AnonIpEnrichment(
...
+    }
+
+    if (ip.startsWith(IPv4MappedAddressPrefix))
+      IPv4MappedAddressPrefix + masked(ip.replace(IPv4MappedAddressPrefix, ""))
+    else masked(ip)
+  }
+
+  private def anonymizeIpV6(ip: String): String =
+    InetAddress
+      .getByName(ip)
```
chuwy: As far as I understand it makes blocking DNS IO call. Can we avoid that?

3-enrich/scala-common-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/registry/AnonIpEnrichmentSpec.scala
```diff
@@ -23,20 +23,35 @@ import org.specs2.matcher.DataTables
  */
 class AnonIpEnrichmentSpec extends Specification with DataTables {
 
-  def is = s2"Anonymizing 0-4 octets across a variety of IP addresses should work $e1"
+  def is = s2"Anonymizing across a variety of IP addresses (v4 and v6) should work $e1"
```
chuwy: Can we come up with a property based test?

E.g. "Given valid ip address and `Octet.N` the final value always contains N `x`s"

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
dilyand: Yeah, I got that. But `split` is one of those weird words that are the same in the past tense, like `put` and `cut`.

3-enrich/scala-common-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/registry/AnonIpEnrichmentSpec.scala
chuwy: Or "Givena valid IP address and Octet - if we replace all `x`s with `1` it should be again a valid IP address"

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
lukeindykiewicz: 🤦‍♂ thanks!

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
lukeindykiewicz: If it's IP, not host name it doesn't do the call. I have commented it after opening PR, https://github.com/snowplow/snowplow/pull/4223#pullrequestreview-302575456

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
```diff
@@ -52,23 +55,30 @@ object AnonIpEnrichment extends ParseableEnrichment {
...
    * @return a configured AnonIpEnrichment instance
    */
-  def parse(config: JValue, schemaKey: SchemaKey): ValidatedNelMessage[AnonIpEnrichment] =
-    isParseable(config, schemaKey).flatMap(conf => {
+  def parse(config: JValue, schemaKey: SchemaKey): ValidatedNelMessage[AnonIpEnrichment] = {
+
+    def extractParam(name: String) =
+      ScalazJson4sUtils.extract[Int](config, "parameters", name)
+
+    isParseable(config, schemaKey).flatMap(_ => {
```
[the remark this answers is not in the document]
lukeindykiewicz: This code is very similar in other enrichments, so I wanted to to stay consistent.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
```diff
@@ -118,23 +167,60 @@ case class AnonIpEnrichment(
...
-              if (octets.id >= All.id - i) "x" else q
-            }
-          }
-          .mkString("."))
-      .orNull
+  val IPv4MappedAddressPrefix = "::FFFF:"
+  val MaskChar                = "x"
+
+  def anonymizeIp(ipOrNull: String): String =
+    Option(ipOrNull).map { ip =>
```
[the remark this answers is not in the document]
lukeindykiewicz: I agree, but this code in applicative style looks terrible, as we discussed offline.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
lukeindykiewicz: `2605.2700.0.3..4713.93e3` will produce `x.x.x.x.x.x.x` for `Six`
`2605.2700.0.3..4713.93e3` will produce `2605.2700.x.x.x.x.x.x` for `Four` 
I don't know if we could anything reasonable about it.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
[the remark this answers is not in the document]
chuwy: Makes sense, it's a nit. But I'd like to make it readable everywhere eventually.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
lukeindykiewicz: ok

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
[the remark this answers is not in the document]
lukeindykiewicz: :+1:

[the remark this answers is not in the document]
lukeindykiewicz: sorry, I removed my previous comment before I saw yours.

3-enrich/scala-common-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/registry/AnonIpEnrichmentSpec.scala
lukeindykiewicz: done

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
chuwy: @benjben do you know if R118 will still have guava as a dependency?

lukeindykiewicz: I added Guava explicitly to `Dependencies`, but it of course was already as some transitive dependency. The versions didn't clash also.

lukeindykiewicz:
Commits squashed.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
chuwy: Just in case - that's a tricky trade-off - we don't want to eventually end up with a guava dependency _just_ for IPs (which we probably won't know). But probably that's still better than potential IO.

3-enrich/scala-common-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/registry/EnrichmentConfigsSpec.scala
lukeindykiewicz: https://github.com/snowplow/iglu-central/issues/1008

lukeindykiewicz:
Example in `3-enrich/config/enrichments/anon_ip.json` updated

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
benjben: I would say that the solution is not to have a workaround for such cases here but rather to fix the upstream to not receive IPv6 with such formats.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
benjben: I can't find a use of guava in SCE for R118. But as long as it's there as a transitive dependency, why not using it.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
```diff
@@ -52,23 +56,30 @@ object AnonIpEnrichment extends ParseableEnrichment {
...
 
 /**
- * How many octets to anonymize?
+ * How many octets (ipv4) to anonymize?
  */
-object AnonOctets extends Enumeration {
+object AnonIPv4Octets extends Enumeration {
 
-  type AnonOctets = Value
+  type AnonIPv4Octets = Value
```
benjben: In the configuration you have `anonOctets` and `anonSegments` and in the code `AnonIPv4Octets` and `AnonIPv6Segments`, could be more consistent.

3-enrich/scala-common-enrich/project/Dependencies.scala
```diff
@@ -50,6 +50,7 @@ object Dependencies {
     val iabClient        = "0.1.0"
     val yauaa            = "5.8"
     val kryo             = "2.24.0"
+    val guava            = "28.1-jre"
```
chuwy: Can we mention that direct dependency is only used for IP representation (just in case somebody will find a better tool for that a decide to get rid of it entirely)?

3-enrich/scala-common-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/registry/AnonIpEnrichmentSpec.scala
```diff
@@ -14,29 +14,101 @@ package com.snowplowanalytics.snowplow.enrich.common
...
+    (for {
+      segNum <- Gen.choose(2, 7)
+      ip     <- Gen.listOfN(segNum, segmentGen)
+    } yield ip.updated(segNum - 1, s":${ip(segNum - 1)}")).map(_.mkString(":"))
+
+  def e2 =
+    forAll(ipv4Gen, octetsNumGen) {
+      case (ip, octetsNum) =>
+        val anon = AnonIpEnrichment(AnonIPv4Octets(octetsNum), AnonIPv6Segments(1)).anonymizeIp(ip)
+        anon.count(_ == 'x') ==== octetsNum
```
chuwy: Are we sure these are assertion-based test?

E.g. usually specs2 immutable test with

```scala
1 must beEqualTo(2)
"ok" must beEqualTo("ok")
```

Won't fail because its value is `ok` and first statement will be simply ignored.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
```diff
@@ -77,64 +88,144 @@ object AnonOctets extends Enumeration {
...
+      Try(GuavaInetAddress.forString(ip))
+        .map {
+          case _: Inet4Address    => anonymizeIpV4(ip)
+          case ipv6: Inet6Address => anonymizeIpV6(ipv6.getHostAddress)
+        }
+        .getOrElse(tryAnonymizingInvalidIp(ip))
+    }.orNull
+
+  private def anonymizeIpV4(ip: String): String = {
+    def masked(ipv4: String) = {
```
benjben: Shouldn't it be `masked` -> `mask`?

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
```diff
@@ -77,64 +88,144 @@ object AnonOctets extends Enumeration {
...
+      val split = ipv4.split("\\.")
+      split
+        .take(AnonIPv4Octets.All.id - ipv4Octets.id)
+        .toList
+        .padTo(split.size, MaskChar)
+        .mkString(".")
+    }
+
+    if (ip.startsWith(IPv4MappedAddressPrefix))
```
benjben: I can't see in which scenario an IPv4 would start with `::FFFF:` ?

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
```diff
@@ -77,64 +88,144 @@ object AnonOctets extends Enumeration {
...
+      IPv4MappedAddressPrefix + masked(ip.replace(IPv4MappedAddressPrefix, ""))
+    else masked(ip)
+  }
+
+  private def anonymizeIpV6(ip: String): String =
```
benjben: `anonymizeIpV4` and `anonymizeIpV6` could be combined into one generic function

3-enrich/scala-common-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/registry/EnrichmentConfigsSpec.scala
```diff
@@ -63,7 +63,26 @@ class EnrichmentConfigsSpec extends Specification with ValidationMatchers with D
...
+
+      val ipAnonJson = parse("""{
+        "enabled": true,
+        "parameters": {
+          "anonOctets": 2,
+          "anonSegments": 3
+        }
+      }""")
+
+      val schemaKey = SchemaKey("com.snowplowanalytics.snowplow", "anon_ip", "jsonschema", "1-0-0")
```
benjben: Should be > `1-0-0`

3-enrich/scala-common-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/registry/AnonIpEnrichmentSpec.scala
lukeindykiewicz: Great catch @chuwy! Thank you. I haven't been using specs2 immutable spec before.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
lukeindykiewicz: I have deliberately left config name as it was before (`anonOctets`, and added in a consistent way `anonSegemnts`), while in code I wanted to describe it in more precise way, so I added `IPv4` and `IPv6` to the names.

3-enrich/scala-common-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/registry/EnrichmentConfigsSpec.scala
lukeindykiewicz: Thanks!

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
lukeindykiewicz: done

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
lukeindykiewicz: Both `java.netInetAddress` and `guava InetAddess` reports ip starting with `::FFFF:` as IPv4.
```
@ InetAddress.getByName("::FFFF:192.168.0.10").isInstanceOf[Inet4Address]
res90: Boolean = true

@ GuavaInetAddress.forString("::FFFF:192.168.0.10").isInstanceOf[Inet4Address]
res91: Boolean = true
```

This usage is rather deprecated, but I think that it's still nice to support it. More info here: https://guava.dev/releases/28.1-jre/api/docs/com/google/common/net/InetAddresses.html

I also added a test for `::192.168.0.1` - this is IPv6 (ipv4 compatible)

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
lukeindykiewicz: Theoretically it could be (I was also thinking about it ;)), but the generic version would need to take 4 params, so I don't think it would make it better actually.

3-enrich/scala-common-enrich/project/Dependencies.scala
lukeindykiewicz: comment added

3-enrich/config/enrichments/anon_ip.json
```diff
@@ -1,13 +1,12 @@
 {
 	"schema": "iglu:com.snowplowanalytics.snowplow/anon_ip/jsonschema/1-0-0",
```
chuwy: Should be `1-0-1`

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/AnonIpEnrichment.scala
benjben: Wow that's good to know! Thanks for the explanation :+1:

3-enrich/config/enrichments/anon_ip.json
lukeindykiewicz: thanks, done

PR #4279, opened 2020-02-05
Update config.hocon.sample

piyushknoldus:
1) Replacing the '->' symbol with '=' , as if we uncomment the settings will result in a runnable error.
Error Details -
[error] (run-main-0) com.typesafe.config.ConfigException$Parse: /home/piyush/Desktop/myconf.conf: 281: Key 'bootstrap.servers -> brokers' may not be followed by token: 'retries' (if you intended 'retries' to be part of a key or string value, try enclosing the key or value in double quotes)
[error] com.typesafe.config.ConfigException$Parse: /home/piyush/Desktop/myconf.conf: 281: Key 'bootstrap.servers -> brokers' may not be followed by token: 'retries' (if you intended 'retries' to be part of a key or string value, try enclosing the key or value in double quotes)

2) Removing the duplicate setting       # retries             = retries, as its been defined and placed with the default value.
Also if once uncommented results in parsing error.

### DIFF
--- a/2-collectors/scala-stream-collector/examples/config.hocon.sample
+++ b/2-collectors/scala-stream-collector/examples/config.hocon.sample
@@ -277,10 +277,9 @@ collector {
       # The kafka producer has a variety of possible configuration options defined at
       # https://kafka.apache.org/documentation/#producerconfigs
       # Some values are set to other values from this config by default:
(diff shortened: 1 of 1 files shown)


benjben:
Thank you for the fix @piyushknoldus !

Issue #3795, opened 2018-06-11
Scala Common Enrich: add validation for contexts added by enrichments


chuwy:
API Request enrichment works on *already* validated contexts and self-described events and does not validate produced contexts, therefore there's always a chance that output data won't correspond to `outputs.schema`.

chuwy:
- [ ] Check same for SQL Query enrichment

benjben:
Hey @chuwy , we talked about this with Customer Squad, that will be in charge of displaying the errors, and we think that this type of error should be a subtype of `EnrichmentFailure`, like `InvalidContext`, because an enrichment that creates a wrong context is a problem with this enrichment. After each enrichment, we would check the context that it added and if it's not valid we create this type of bad row and stop the processing. WDYT?

chuwy:
Hey @benjben, sorry missed your last message. As we discussed in internal RFC doc yesterday, there are many more oppostunities for loaders to fail. Biggest ones are:

* Unavailable (or wrong version of) Iglu Registry. Loaders have their own requirements for registries
* Unavailable DB. BQ failed inserts exist because BigQuery was unavailable
* Bug in a shredding/casting algorithm
* Improperly versioned schema

chuwy:
Possible duplicate: https://github.com/snowplow/snowplow/issues/2639

PR #4287, opened 2020-02-29
Scala Common Enrich: add validation for contexts added by enrichments

benjben:
Related to #3795

`Shredder` became `IgluUtils` and was rewritten. I took the opportunity to add decent unit tests to the validation of unstructured events and input contexts as well. The unit tests for `IgluUtils` are:
```
[info] IgluUtilsSpec
[info] extractAndValidateUnstructEvent should
[info]   + return None if unstruct_event field is empty
[info]   + return a SchemaViolation.NotJson if unstruct_event does not contain a properly formatted JSON string
[info]   + return a SchemaViolation.NotIglu if unstruct_event contains a properly formatted JSON string that is not self-describing
[info]   + return a SchemaViolation.CriterionMismatch if unstruct_event contains a self-describing JSON but not with the expected schema for unstructured events
[info]   + return a SchemaViolation.NotJson if the JSON in .data is not a JSON
[info]   + return a SchemaViolation.IgluError wrapping a ValidationError if the JSON in .data is not self-describing
[info]   + return a SchemaViolation.IgluError wrapping a ValidationError if the JSON in .data is not a valid SDJ
[info]   + return a SchemaViolation.IgluError wrapping a ResolutionError if the schema of the SDJ in .data can't be resolved
[info]   + return the extracted unstructured event if .data is a valid SDJ
[info] extractAndValidateInputContexts should
[info]   + return Nil if contexts field is empty
[info]   + return a SchemaViolation.NotJson if .contexts does not contain a properly formatted JSON string
[info]   + return a SchemaViolation.NotIglu if .contexts contains a properly formatted JSON string that is not self-describing
[info]   + return a SchemaViolation.CriterionMismatch if .contexts contains a self-describing JSON but not with the right schema
[info]   + return a SchemaViolation.IgluError wrapping a ValidationError if .data does not contain an array of JSON objects
[info]   + return a SchemaViolation.IgluError wrapping a ValidationError if .data contains one invalid context
[info]   + return a SchemaViolation.IgluError wrapping a ResolutionError if .data contains one context whose schema can't be resolved
[info]   + return 2 expected failures for 2 invalid contexts
[info]   + return an expected failure if one context is valid and the other invalid
[info]   + return the extracted SDJs for 2 valid input contexts
[info] validateEnrichmentsContexts should
[info]   + return an expected failure for one invalid context
[info]   + return 2 expected failures for 2 invalid contexts
[info]   + return an expected failure for 1 valid context and one invalid
[info]   + not return any error for 2 valid contexts
```

I refactored `EnrichmentManager.enrichEvent()` so that only one bad row can be emitted per event (in R118 both an `EnrichmentFailures` and a `SchemaViolations` could be emitted at the same time for the same event). The new logic is:
1. Validate unstructured event and contexts. If not valid, short-circuit and emit `SchemaViolations`
2. Run the enrichments. If problem(s), short-circuit and emit `EnrichmentFailures`
3. Check that all the contexts added by the enrichments are valid. If not, emit `EnrichmentFailures` wrapping an `IgluError`.

`EnrichmentManager.enrichEvent()` was not unit tested so I created it to test the new logic:
```
[info] EnrichmentManagerSpec
[info] enrichEvent should
[info]   + emit a SchemaViolations wrapping 2 expected errors if the input event contains 2 invalid contexts
[info]   + emit a SchemaViolations wrapping 1 expected error if the input unstructured event is invalid
[info]   + emit a SchemaViolations wrapping 3 expected errors if the input event contains 2 invalid contexts and 1 invalid unstructured event
[info]   + emit an EnrichmentFailures wrapping an expected error (Simple here) if one of the enrichment fails (JS enrichment here)
[info]   + emit an EnrichmentFailures wrapping an IgluError if one of the contexts added by the enrichments is invalid (using JS enrichment)
[info]   + emit an EnrichedEvent if the input event contains a valid context and a valid unstructured event
```

I removed [EnrichmentStageIssue](https://github.com/snowplow-incubator/snowplow-badrows/blob/master/src/main/scala/com.snowplowanalytics.snowplow.badrows/FailureDetails.scala#L105) and replaced it by either `SchemaViolation` or `EnrichmentFailure` everywhere, to always be as specific as we can be.

I also updated `SchemaEnrichment` because with the new workflow it can access the result of `IgluUtils.extractAndValidateUnstructEvent` and does not need to extract the unstructured event again.

Schemas for unit tests are in `src/test/resources/iglu-schemas/`.

As a reminder, a [SchemaViolations](https://github.com/snowplow-incubator/snowplow-badrows/blob/master/src/main/scala/com.snowplowanalytics.snowplow.badrows/BadRow.scala#L127) bad row can contain several [FailureDetails.SchemaViolation](https://github.com/snowplow-incubator/snowplow-badrows/blob/master/src/main/scala/com.snowplowanalytics.snowplow.badrows/FailureDetails.scala#L107), and an [EnrichmentFailures](https://github.com/snowplow-incubator/snowplow-badrows/blob/master/src/main/scala/com.snowplowanalytics.snowplow.badrows/BadRow.scala#L140) bad row can contain several [FailureDetails.EnrichmentFailure](https://github.com/snowplow-incubator/snowplow-badrows/blob/master/src/main/scala/com.snowplowanalytics.snowplow.badrows/FailureDetails.scala#L130).

Important files of the PR:
- `EnrichmentManager` + `EnrichmentManagerSpec`
- `IgluUtils` + `IgluUtilsSpec`
- `SchemaEnrichment` + `SchemaEnrichmentSpec`


3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/IgluUtils.scala
```diff
@@ -0,0 +1,206 @@
...
+          unstruct <- extractInputData(rawUnstructEvent, field, criterion, client)
+          // Parse Json unstructured event as SelfDescribingData[Json]
+          unstructSDJ <- SelfDescribingData
+            .parse(unstruct)
+            .leftMap(e => FailureDetails.SchemaViolation.NotIglu(unstruct, e))
+            .toEitherT[F]
+          // Check SelfDescribingData[Json] of unstructured event
+          _ <- check(client, unstructSDJ)
+            .leftMap(e => FailureDetails.SchemaViolation.IgluError(e._1, e._2))
+            .leftWiden[FailureDetails.SchemaViolation]
```
benjben: Could someone of @snowplow/com-snowplowanalytics-engineering-datacapability explain to me why I need `leftWiden` here but not on line 63 please?

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentManager.scala
```diff
@@ -72,233 +74,288 @@ object EnrichmentManager {
...
+            derivedTstamp.toValidatedNel,
+            iabContext.toValidated,
+            uaUtils.toValidatedNel,
+            uaParser.toValidatedNel,
+            refererUri.toValidatedNel,
+            pageQsMap.toValidatedNel,
+            crossDomain.toValidatedNel,
+            extractSchema.toValidatedNel,
+            jsScript.toValidatedNel
+          ).mapN((_, _, _, _, _, _, _, _, _, _, _, _, _) => ())
```
chuwy, line 258: I just realized that it can be:

```suggestion
          List(
            collectorTstamp.toValidatedNel,
            useragent.toValidatedNel,
            collectorVersionSet.toValidatedNel,
            pageUri.toValidatedNel,
            derivedTstamp.toValidatedNel,
            iabContext.toValidated,
            uaUtils.toValidatedNel,
            uaParser.toValidatedNel,
            refererUri.toValidatedNel,
            pageQsMap.toValidatedNel,
            crossDomain.toValidatedNel,
            extractSchema.toValidatedNel,
            jsScript.toValidatedNel
          ).sequence_
```

So we don't need to care about arity.

3-enrich/scala-common-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/utils/IgluUtilsSpec.scala
```diff
@@ -0,0 +1,422 @@
...
+    """{"schema":"iglu:com.snowplowanalytics.snowplow/foo/jsonschema/1-0-0", "data": {}}"""
+
+  "extractAndValidateUnstructEvent" should {
+    "return None if unstruct_event field is empty" >> {
+      IgluUtils
+        .extractAndValidateUnstructEvent(new EnrichedEvent, SpecHelpers.client)
+        .value must beValid(None)
+    }
+
+    "return a SchemaViolation.NotJson if unstruct_event does not contain a properly formatted JSON string" >> {
```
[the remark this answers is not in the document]
benjben: I hesitated about this, but given that the function has not much complexity, I decided to test it through the public functions `extractAndValidateUnstructEvent` and `extractAndValidateInputContexts`, as suggested in  [this thread](https://stackoverflow.com/questions/105007/should-i-test-private-methods-or-only-public-ones).

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/IgluUtils.scala
```diff
@@ -0,0 +1,219 @@
...
+
+  /** Extract unstructured event from event and validate against its schema
+   *  @param event Snowplow event from which to extract unstructured event (in String)
+   *  @param client Iglu client used for SDJ validation
+   *  @param field Name of the field containing the unstructured event, to put in the bad row
+   *               in case of failure
+   *  @param criterion Expected schema for the JSON containing the unstructured event
+   *  @return Valid unstructured event if the input event has one
+   */
+  def extractAndValidateUnstructEvent[F[_]: Monad: RegistryLookup: Clock](
```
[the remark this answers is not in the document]
chuwy: No, I'd go with @benjben's approach. Its really easy to confuse them in many places. `Option` and `List` give us just enough protection.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentManager.scala
```diff
@@ -72,233 +74,288 @@ object EnrichmentManager {
...
+    val firstPassTransform = event.transform(sourceMap, firstPassTransformMap)
+    val secondPassTransform = event.transform(sourceMap, secondPassTransformMap)
 
-    // 2a. Failable enrichments which don't need the payload
+    val transform = EitherT(
+      for {
+        first <- Monad[F].pure(firstPassTransform)
+        second <- Monad[F].pure(secondPassTransform)
+      } yield (first, second)
+        .mapN { (_, _) =>
```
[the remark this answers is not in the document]
benjben: `.mapN()` makes you save one step as it's kind of equivalent of a `combine` and then `map`, although this is not `combine` but `traverse`. You can find more details [here](https://typelevel.org/cats/typeclasses/applicative.html#traverse).

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/IgluUtils.scala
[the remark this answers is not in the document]
benjben: I think that having `extractAndValidateUnstructEvent` returning a `List` would be misleading as it would let us think that more than 1 element could be returned, whereas it will always be maximum one.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/Shredder.scala
```diff
@@ -1,217 +0,0 @@
...
-        Monad[F].pure(List.empty[SelfDescribingData[Json]].asRight)
-    }
-
-  /**
-   * Extract list of derived contexts out of string and validate each against its schema
-   * @param event The Snowplow enriched event to extract custom context JSONs from
-   * @param client The Iglu client used for schema lookup
-   * @return validated list of pairs consist of schema and node
-   */
-  def extractAndValidateDerivedContexts[F[_]: Monad: RegistryLookup: Clock](
```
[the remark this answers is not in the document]
benjben: Exactly

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/IgluUtils.scala
```diff
@@ -0,0 +1,219 @@
...
+            .leftMap {
+              case (schemaKey, clientError) =>
+                FailureDetails.SchemaViolation.IgluError(schemaKey, clientError)
+            }
+            .leftWiden[FailureDetails.SchemaViolation]
+        } yield unstructSDJ.some
+      case None =>
+        EitherT.rightT[F, FailureDetails.SchemaViolation](none[SelfDescribingData[Json]])
+    }).value
+      .map(_.toValidated)
```
chuwy, line 75:
```suggestion
    }).toValidated
```

Any `EitherT`'s `value.map(_.toValidated)` can be replaced with just with `toValidated`

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/IgluUtils.scala
```diff
@@ -0,0 +1,219 @@
...
+              .sequence
+              .map(_.sequence.toEither)
+          )
+        } yield contextsSDJ
+      case None =>
+        EitherT.rightT[F, NonEmptyList[FailureDetails.SchemaViolation]](
+          List.empty[SelfDescribingData[Json]]
+        )
+    }).value
+      .map(_.toValidated)
```
chuwy, line 112:
```suggestion
    }).toValidated
```

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/IgluUtils.scala
```diff
@@ -0,0 +1,219 @@
...
+            val enrichmentInfo =
+              FailureDetails.EnrichmentInformation(schemaKey, "enrichments-contexts-validation")
+            FailureDetails.EnrichmentFailure(
+              enrichmentInfo.some,
+              FailureDetails.EnrichmentFailureMessage.IgluError(schemaKey, clientError)
+            )
+        }
+      )
+      .value
+      .map(_.toValidated)
```
chuwy, line 136:
```suggestion
      .toValidated
```

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentManager.scala
```diff
@@ -72,233 +74,288 @@ object EnrichmentManager {
...
+      _ <- EitherT(for {
+        valid <- IgluUtils.validateEnrichmentsContexts[F](client, enrichmentsContexts)
+      } yield valid.leftMap { enrichmentFailures =>
+        buildEnrichmentFailuresBadRow(
+          enrichmentFailures,
+          EnrichedEvent.toPartiallyEnrichedEvent(event),
+          RawEvent.toRawEvent(raw),
+          processor
+        )
+      }.toEither)
```
chuwy, line 341:
```suggestion
      _ <- EitherT(IgluUtils.validateEnrichmentsContexts[F](client, enrichmentsContexts).map(_.toEither))
        .leftMap { enrichmentFailures =>
          buildEnrichmentFailuresBadRow(
            enrichmentFailures,
            EnrichedEvent.toPartiallyEnrichedEvent(event),
            RawEvent.toRawEvent(raw),
            processor
          )
        }
```

Just to get rid of `for`-comprehension.

3-enrich/scala-common-enrich/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/EnrichmentManagerSpec.scala
```diff
@@ -0,0 +1,390 @@
...
+  val context = CollectorPayload.Context(
+    DateTime.parse("2013-08-29T00:18:48.000+00:00").some,
+    "37.157.33.123".some,
+    None,
+    None,
+    Nil,
+    None
+  )
+
+  "enrichEvent" should {
```
chuwy, line 55: 👍

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentManager.scala
```diff
@@ -72,233 +74,288 @@ object EnrichmentManager {
...
+            )
+          }
+          .toEither
+      )
 
-    // The load fails if the collector version is not set
-    val collectorVersionSet: Either[FailureDetails.EnrichmentStageIssue, Unit] =
-      getCollectorVersionSet(event)
+      ////////////////////////////////////////////////////////
+      // 3. Run enrichments and check that they are successful
```
chuwy, line 138: Can we move the whole 3 into a separate function, something like:

```scala
def runEnrichments(event: EnrichedEvent, inputContexts: List[SelfDescribingData[Json]]): EitherT[F, BadRow, List[SelfDescribingData[Json]]] = ???
```

And then this whole 300LoC function would look like:

```scala
for {
  _                              <- transform
  (inputContexts, unstructEvent) <- IgluUtils.extractTrackedJsons(event, client)
  derivedContexts                <- runEnrichments(event, inputContexts)
  _                              <- IgluUtils.validateEnrichmentsContexts[F](client, derivedContexts)
  _                              <- anonymize(event)
} yield event
```

Which is basically an essence of the enrichment.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/IgluUtils.scala
```diff
@@ -0,0 +1,219 @@
...
+        )
+    }).value
+      .map(_.toValidated)
+
+  /** Validate each context added by the enrichments against its schema
+   *  @param client Iglu client used for SDJ validation
+   *  @param sdjs List of enrichments contexts to be added to the enriched event
+   *  @return Unit if all the contexts are valid
+   */
+  def validateEnrichmentsContexts[F[_]: Monad: RegistryLookup: Clock](
```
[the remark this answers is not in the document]
benjben: Both `extractAndValidateUnstructEvent` and `extractAndValidateInputContexts` take an `EnrichedEvent` as input, but as they don't do the same thing the other input arguments and return type are different.

Also, they don't return a `Validated` but a `F[Validated]`, so `andThen` can't be used to chain them directly.

chuwy:
> May I ask why we'd prefer Either over Validated? Is it to use Scala lib wherever we can instead of an external library?

Yes! My hope is that if at some point people outside Snowplow will start using SCE, the entry-point signature will be a bit more familiar. Though this is really-really minor.

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/EtlPipeline.scala
```diff
@@ -74,76 +71,17 @@ object EtlPipeline {
       events <- EitherT(
         adapterRegistry
           .toRawEvents(payload, client, processor)
-          .map(_.toEither.leftMap(sd => NonEmptyList.one(sd)))
+          .map(_.toEither.leftMap(badRow => NonEmptyList.one(badRow)))
```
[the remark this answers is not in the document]
benjben: Seems that we can't go directly from a `Validated[E, A]` to a `Either[NonEmptyList[E],A]`

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentManager.scala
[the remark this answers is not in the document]
benjben: > if firstPassTransform fails - we don't get results of secondPassTransform

Actually I don't get the problem with `mapN`.

```scala
scala> val a: ValidatedNel[String, Int] = Validated.Invalid(NonEmptyList.one("a"))
a: cats.data.ValidatedNel[String,Int] = Invalid(NonEmptyList(a))

scala> val b: ValidatedNel[String, Int] = Validated.Invalid(NonEmptyList.one("b"))
b: cats.data.ValidatedNel[String,Int] = Invalid(NonEmptyList(b))

scala> (a, b).mapN((aa, bb) => (aa, bb))
res0: cats.data.ValidatedNel[String,(Int, Int)] = Invalid(NonEmptyList(a, b))
```

The 2 errors are returned, not only the first one. Am I missing something?

benjben:
Hi @dilyand and @chuwy , I addressed your feedback (in https://github.com/snowplow/snowplow/pull/4287/commits/7a73d3156a826babd4112631c37377183f5eb71f). Mainly it includes:

- Move the `TransformMap` stuff to its own file (`Transform.scala`). It contains the mapping/transforming of the input fields of the raw event to fields of the enriched event
- Create function `EnrichmentManager.enrichEvent()` to run all the enrichments
- Create function `IgluUtils.extractAndValidateInputJsons()` to extract and validate the unstructured event and the input contexts at once, in order to reduce the logic in `EnrichmentManager`
- `EnrichmentManager.enrichEvent()` is now just a for - comprehension loop
- Update unit tests regarding ^

3-enrich/scala-common-enrich/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/Transform.scala
```diff
@@ -0,0 +1,135 @@
+package com.snowplowanalytics.snowplow.enrich.common
```
chuwy, line 1: Missing license header.

PR #4301, opened 2020-03-17
Clojure Collector: add a deprecation note (closes #4264)

### LINKED ISSUE #4264: Clojure Collector: deprecate
There will ve a discourse post as well

benjben:


2-collectors/clojure-collector/README.md
```diff
@@ -1,45 +1,3 @@
 # Clojure Collector
 
-## Introduction
```
chuwy: I think maybe we should remove the directory altogether?

benjben:
Handled by #4337

PR #4307, opened 2020-03-26
Allow Stream Enrich to download data from private S3 or GCS

oguzhanunlu:


3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/utils.scala
```diff
@@ -49,4 +70,98 @@ object utils {
...
+        case AWSCredentials(a, s) if isIam(a) && isIam(s) =>
+          "accessKey and secretKey must both be set to 'iam' or neither".asLeft
+        case AWSCredentials(a, s) if isEnv(a) && isEnv(s) =>
+          new EnvironmentVariableCredentialsProvider().asRight
+        case AWSCredentials(a, s) if isEnv(a) || isEnv(s) =>
+          "accessKey and secretKey must both be set to 'env' or neither".asLeft
+        case AWSCredentials(a, s) =>
+          new AWSStaticCredentialsProvider(new BasicAWSCredentials(a, s)).asRight
+      }
+    } yield provider
```
benjben: Could be put directly in `case c: AWSCredentials => c.asRight`

3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/utils.scala
```diff
@@ -49,4 +70,98 @@ object utils {
...
+      credsPath = gcpCreds.creds
+      googleCreds <- {
+        if (Files.isRegularFile(Paths.get(credsPath)))
+          GoogleCredentials
+            .fromStream(new FileInputStream(credsPath))
+            .createScoped("https://www.googleapis.com/auth/cloud-platform")
+            .asRight
+        else "Provided Google Credentials Path isn't valid".asLeft
+      }
+    } yield googleCreds
```
benjben: Could be put directly in `case c: GCPCredentials => c.asRight`

3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/utils.scala
```diff
@@ -49,4 +70,98 @@ object utils {
...
+        case _: AWSCredentials => "AWS credentials provided".asLeft
+        case c: GCPCredentials => c.asRight
+      }
+      credsPath = gcpCreds.creds
+      googleCreds <- {
+        if (Files.isRegularFile(Paths.get(credsPath)))
+          GoogleCredentials
+            .fromStream(new FileInputStream(credsPath))
+            .createScoped("https://www.googleapis.com/auth/cloud-platform")
+            .asRight
```
benjben: Can't this go wrong?

3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/utils.scala
```diff
@@ -49,4 +70,98 @@ object utils {
...
+   * @param uri The URI to reconstruct into a signed S3 URL
+   * @param targetFile The file object to write to
+   * @param provider necessary credentials to download from S3
+   * @return the download result
+   */
+  def downloadFromS3(
+    provider: AWSCredentialsProvider,
+    uri: URI,
+    targetFile: File
+  ): Either[Throwable, Int] = {
```
benjben: Could be `Either[Throwable, Int]`, we don't need to return `Right(0)`, if it's a `Right` we know that it was successful

3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/utils.scala
```diff
@@ -49,4 +70,98 @@ object utils {
...
+      s3Client.getObject(new GetObjectRequest(bucketName, key), targetFile)
+      0
+    }
+  }
+
+  def downloadFromGCS(
+    creds: GoogleCredentials,
+    uri: URI,
+    targetFile: File
+  ): Either[Throwable, Int] = {
```
benjben: Same

3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/utils.scala
```diff
@@ -49,4 +70,98 @@ object utils {
...
+   * @param targetFile The file object to write to
+   * @param provider necessary credentials to download from S3
+   * @return the download result
+   */
+  def downloadFromS3(
+    provider: AWSCredentialsProvider,
+    uri: URI,
+    targetFile: File
+  ): Either[Throwable, Int] = {
+    val s3Client = AmazonS3ClientBuilder.standard().withCredentials(provider).build()
```
benjben: Can't the building of the client go wrong?

3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/utils.scala
```diff
@@ -49,4 +70,98 @@ object utils {
...
+      0
+    }
+  }
+
+  def downloadFromGCS(
+    creds: GoogleCredentials,
+    uri: URI,
+    targetFile: File
+  ): Either[Throwable, Int] = {
+    val storage = StorageOptions.newBuilder().setCredentials(creds).build().getService
```
benjben: Can't the creation of `storage` go wrong?

3-enrich/stream-enrich/kinesis/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/KinesisEnrich.scala
```diff
@@ -61,12 +56,12 @@ object KinesisEnrich extends Enrich {
...
-        case c: Kinesis => c.aws.asRight
+        case c: Kinesis => (c.aws, c.gcp.fold[Credentials](NoCredentials)(identity)).asRight
         case _ => "Configured source/sink is not Kinesis".asLeft
       }
-      client <- parseClient(resolverArg)(creds)
-      enrichmentsConf <- parseEnrichmentRegistry(enrichmentsArg, client)(creds)
-      _ <- cacheFiles(enrichmentsConf, forceDownload)(creds)
+      client <- parseClient(resolverArg)(creds._1)
+      enrichmentsConf <- parseEnrichmentRegistry(enrichmentsArg, client)(creds._1)
+      _ <- cacheFiles(enrichmentsConf, forceDownload)(creds._1, creds._2)
```
benjben: Could be nice to create a case class to remove all these `._1` and `._2`

3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/Enrich.scala
```diff
@@ -63,9 +63,14 @@ trait Enrich {
     val trackerSource: Either[String, (Option[Tracker[Id]], Source)] = for {
       config <- parseConfig(args)
       (enrichConfig, resolverArg, enrichmentsArg, forceDownload) = config
+      sourceSink = enrichConfig.streams.sourceSink.asInstanceOf[SourceSinkAgnosticConfig]
+      credentials <- (
+        sourceSink.aws.fold[Credentials](NoCredentials)(identity),
+        sourceSink.gcp.fold[Credentials](NoCredentials)(identity)
+      ).asRight
```
benjben: Would be nice to add a unit test that checks that `credentials` gets the expected value depending on the configuration

3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/utils.scala
oguzhanunlu: The reason we return successful exit code after a potentially exception-throwing line actually comes from the first branch of `download()`. Please check the method and you'll see that in case uri scheme is `http(s)`, we use `!`:

```
def ! : Int // returns exit code
```

Both AWS and GCP SDKs' relevant methods return different types and both aren't `Int`.

Upstream in the codebase, `cacheFiles` method [uses](https://github.com/snowplow/snowplow/blob/master/3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/Enrich.scala#L285-L288) exit code semantic to interpret download failures.

I want to double check if you still think we should get rid of exit code usage altogether, which also means to update how `cacheFiles` interpret failure. Currently it doesn't make any use of failures (`Left` values) received from credential or download methods. @benjben

benjben: Thanks for the details @oguzhanunlu.

> I say we get rid of exit code semantics and make use of thrown exceptions to interpret download failures, wdyt?

Totally with you!

oguzhanunlu:
@benjben I addressed your feedback in separate commits, please let me know if they meet your feedback, I'll squash once we agree with the changes.

3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/model.scala
```diff
@@ -29,6 +29,8 @@ object model {
   final case class AWSCredentials(accessKey: String, secretKey: String) extends Credentials
   final case class GCPCredentials(creds: String) extends Credentials
 
+  case class MultiCloudCredentials(aws: Credentials, gcp: Credentials)
```
benjben, line 32: Could be final

3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/Enrich.scala
```diff
@@ -258,7 +258,7 @@ a  * @param creds optionally necessary credentials to download the resolver
     uri: URI,
     targetFile: File
   )(
```
benjben, line 257: Why not passing the 4 parameters altogether?

3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/utils.scala
```diff
@@ -135,33 +137,31 @@ object utils {
...
     Either.catchNonFatal {
+      val s3Client = AmazonS3ClientBuilder.standard().withCredentials(provider).build()
+      val bucketName = uri.getHost
+      val key = uri.getPath match { // Need to remove leading '/'
+        case s if s.length > 0 && s.charAt(0) == '/' => s.substring(1)
+        case s => s
+      }
       s3Client.getObject(new GetObjectRequest(bucketName, key), targetFile)
       0
     }
```
benjben: We could have more fine-grained error control with a `for {} yield` and individual `Either` with `.leftMap` rather than a global `Either.catchNonFatal {` with everything inside and return `Either[String, Int]` with a clear error message rather than a `Throwable`.

3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/utils.scala
```diff
@@ -135,33 +137,31 @@ object utils {
...
     Either.catchNonFatal {
+      val storage = StorageOptions.newBuilder().setCredentials(creds).build().getService
+      val bucketName = uri.getHost
+      val key = uri.getPath match { // Need to remove leading '/'
+        case s if s.length > 0 && s.charAt(0) == '/' => s.substring(1)
+        case s => s
+      }
       storage.get(BlobId.of(bucketName, key)).downloadTo(targetFile.toPath)
       0
     }
```
benjben: Same

3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/Enrich.scala
```diff
@@ -260,7 +260,7 @@ a  * @param creds optionally necessary credentials to download the resolver
...
     awsCreds: Credentials,
     gcpCreds: Credentials
-  ): Either[String, Int] =
+  ): Either[String, Any] =
```
benjben, line 260:
```suggestion
  ): Either[String, Unit] =
```

3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/Enrich.scala
```diff
@@ -290,7 +290,7 @@ a  * @param creds optionally necessary credentials to download the resolver
   )(
     awsCreds: Credentials,
     gcpCreds: Credentials
-  ): Either[String, List[Int]] = {
+  ): EitherS[List[Any]] = {
```
benjben, line 290:
```suggestion
  ): EitherS[List[Unit]] = {
```

3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/Enrich.scala
```diff
@@ -303,14 +303,13 @@ a  * @param creds optionally necessary credentials to download the resolver
...
       case (cleanURI, targetFile) =>
-        download(cleanURI, targetFile)(awsCreds, gcpCreds).flatMap {
-          case i if i != 0 => s"Attempt to download $cleanURI to $targetFile failed".asLeft
-          case o => o.asRight
+        download(cleanURI, targetFile)(awsCreds, gcpCreds).leftMap { err =>
+          s"Attempt to download $cleanURI to $targetFile failed: $err"
         }
     }
-    downloadedFiles.sequence[EitherS, Int]
+    downloadedFiles.sequence[EitherS, Any]
```
benjben, line 309: You don't need to create `downloadedFiles`, you could have directly
```scala
filteredFiles
  .map { }
  .sequence[EitherS, Unit]
```

benjben: Given that you care only about the errors you can use `.sequence_`

3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/utils.scala
```diff
@@ -146,14 +134,13 @@ object utils {
         case s => s
       }
       s3Client.getObject(new GetObjectRequest(bucketName, key), targetFile)
```
benjben, line 150:
```suggestion
      s3Client.getObject(new GetObjectRequest(bucketName, key), targetFile)
      ()
```

3-enrich/stream-enrich/core/src/test/scala/com.snowplowanalytics.snowplow.enrich.stream/UtilsSpec.scala
```diff
@@ -37,4 +44,25 @@ class UtilsSpec extends Specification {
       utils.emitPii(SpecHelpers.enrichmentRegistry) must_== true
     }
   }
+
+  "extractCredentials" should {
```
benjben, line 48: Great!!

3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/Enrich.scala
```diff
@@ -253,22 +253,24 @@ a  * @param creds optionally necessary credentials to download the resolver
...
     awsCreds: Credentials,
     gcpCreds: Credentials
-  ): Either[String, Any] =
+  ): Either[String, Unit] =
     uri.getScheme match {
-      case "http" | "https" => (uri.toURL #> targetFile).!.asRight
+      case "http" | "https" =>
+        val exitCode = (uri.toURL #> targetFile).!
+        if (exitCode == 0) ().asRight
+        else s"http(s) download failed with exit code $exitCode".asLeft
```
benjben, line 264: You could use directly pattern matching
```scala
(uri.toURL #> targetFile).! match {
  case 0 => ().asRight
  case exitCode =>
    s"http(s) download failed with exit code $exitCode".asLeft
```

3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/Enrich.scala
```diff
@@ -283,31 +285,30 @@ a  * @param creds optionally necessary credentials to download the resolver
...
+        case (_, targetFile) =>
+          forceDownload || targetFile.length == 0L
+      }
+      .map {
+        case (cleanURI, targetFile) =>
+          download(cleanURI, targetFile, awsCreds, gcpCreds).leftMap { err =>
+            s"Attempt to download $cleanURI to $targetFile failed: $err"
+          }
+      }
+      .sequence_
```
benjben, line 311: :+1:

3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/utils.scala
```diff
@@ -139,31 +139,39 @@ object utils {
...
+    for {
+      s3Client <- Either
+        .catchNonFatal(AmazonS3ClientBuilder.standard().withCredentials(provider).build())
+        .leftMap(_.getMessage)
+      bucketName = uri.getHost
+      key = extractObjectKey(uri)
+      _ <- Either
+        .catchNonFatal(s3Client.getObject(new GetObjectRequest(bucketName, key), targetFile))
+        .leftMap(_.getMessage)
+    } yield ()
```
benjben, line 152: :+1:

3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/utils.scala
```diff
@@ -139,31 +139,39 @@ object utils {
...
+      _ <- Either
+        .catchNonFatal(storage.get(BlobId.of(bucketName, key)).downloadTo(targetFile.toPath))
+        .leftMap(_.getMessage)
+    } yield ()
+
+  def extractObjectKey(uri: URI): String = {
+    val path = uri.getPath
+    if (path.length > 0 && path.charAt(0) == '/') path.substring(1)
+    else path
+  }
```
benjben, line 175:
```suggestion
  def extractObjectKey(uri: URI): String = uri.getPath match {
    case path if (path.length > 0 && path.charAt(0) == '/') => path.substring(1)
    case path => path
  }
```

3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/model.scala
```diff
@@ -27,6 +27,9 @@ object model {
   sealed trait Credentials
   case object NoCredentials extends Credentials
   final case class AWSCredentials(accessKey: String, secretKey: String) extends Credentials
+  final case class GCPCredentials(creds: String) extends Credentials
+
+  final case class MultiCloudCredentials(aws: Credentials, gcp: Credentials)
```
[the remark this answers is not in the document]
oguzhanunlu: `MultiCloudCredentials` doesn't extend `Credentials` so that shouldn't be possible.

3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/Enrich.scala
```diff
@@ -245,50 +247,69 @@ a  * @param creds optionally necessary credentials to download the resolver
...
+   * @param awsCreds optionally necessary credentials to download the file
+   * @param gcpCreds optionally necessary credentials to download the file
    * @return the return code of the downloading command
    */
-  def download(uri: URI, targetFile: File)(implicit creds: Credentials): Either[String, Int]
-  val httpDownloader = (uri: URI, targetFile: File) =>
+  def download(
+    uri: URI,
+    targetFile: File,
+    awsCreds: Credentials,
```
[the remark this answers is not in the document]
oguzhanunlu: As we talked, naming is confusing in my PR and I'll fix it but as I wrote below
> MultiCloudCredentials doesn't extend Credentials

so both or either can't be multi-cloud

3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/Enrich.scala
```diff
@@ -63,9 +63,11 @@ trait Enrich {
     val trackerSource: Either[String, (Option[Tracker[Id]], Source)] = for {
       config <- parseConfig(args)
       (enrichConfig, resolverArg, enrichmentsArg, forceDownload) = config
+      sourceSink = enrichConfig.streams.sourceSink.asInstanceOf[SourceSinkAgnosticConfig]
```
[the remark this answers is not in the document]
oguzhanunlu: we talked about it but I note for other reviewers, I'll get rid of reflection here.

3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/model.scala
```diff
@@ -44,10 +47,16 @@ object model {
     partitionKey: String
   )
   final case class KinesisBackoffPolicyConfig(minBackoff: Long, maxBackoff: Long)
-  sealed trait SourceSinkConfig
+  sealed trait SourceSinkConfig {
```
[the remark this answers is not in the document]
oguzhanunlu: Agnostic config doesn't depend on a cloud platform and used for Kafka, NSQ and Stdin whereas SourceSinkConfig is used for Kinesis, hence it isn't platform agnostic. I'll fix naming and add scaladocs and see if I can further improve without touching configuration.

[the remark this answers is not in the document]
oguzhanunlu: > Base source config has GCP for some reasons.

that is because stream-enrich-kinesis might run an enrichment where enrichment requires a file hosted privately in GCS and stream-enrich-kinesis depends on SourceSinkConfig, not SourceSink**Agnostic**Config, since it is platform dependent and not agnostic

Repeated above but I'm just writing to answer the question for all reviewers, I'll fix naming hopefully and add scaladocs to ease understanding

3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/model.scala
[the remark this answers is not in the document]
oguzhanunlu: I must admit that `MultiCloudCredentials` sounds like it is a type of `Credentials` but it isn't. Confusing and misleading.

oguzhanunlu:
> The configuration need some serious rework
I definitely agree with you @chuwy , stream enrich models have to depend on [the way](https://pureconfig.github.io/docs/overriding-behavior-for-sealed-families.html) pureconfig works.

It is then a natural result that whole codebase is shaped and developed based on the model, which constraints the maintainer.

oguzhanunlu:
Hi @chuwy , I tried to address your written & verbal feedback in a separate commit. Please let me know how it looks.

3-enrich/stream-enrich/core/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/model.scala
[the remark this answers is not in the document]
oguzhanunlu: For clarity, I put an explanation last week just to answer the question, these comments are invalid after addressing Anton's feedback, since both types are removed.

PR #4308, opened 2020-03-31
Provide a working example for referer_parser 2.0.0

mleuthold:

### DIFF
--- a/3-enrich/config/enrichments/referer_parser.json
+++ b/3-enrich/config/enrichments/referer_parser.json
@@ -1,5 +1,5 @@
 {
-	"schema": "iglu:com.snowplowanalytics.snowplow/referer_parser/jsonschema/1-0-0",
+	"schema": "iglu:com.snowplowanalytics.snowplow/referer_parser/jsonschema/2-0-0",
 
 	"data": {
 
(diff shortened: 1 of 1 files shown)


mleuthold:
I want to update the enricher component from version 0.20.0 to 1.0.0, i.e. use the provided Docker image from here:
https://hub.docker.com/r/snowplow/stream-enrich-kafka

However, error messages told me, I also need to update this enrichment from 1.0.0 to 2.0.0. But there is no working example, how this enrichment should be configured. Therefore I'm providing it myself to you.

mleuthold:
I would also like to update the wiki page here with an example for referer_parser 2.0.0:
https://github.com/snowplow/snowplow/wiki/referer-parser-enrichment

mleuthold:
The YAML version of the referer-parser file is not working. I had to use the JSON version.

This is the error, when using the YAML version:
` An error occured: expected json value got '# ####...' (line 1, column 1) `

benjben:
Thank you for the contribution @mleuthold , adding it to R119 which will be next Snowplow release.

benjben:
Hi @mleuthold ,

> I would also like to update the wiki page here with an example for referer_parser 2.0.0:
> https://github.com/snowplow/snowplow/wiki/referer-parser-enrichment

We're currently migrating all our documentation to https://docs.snowplowanalytics.com, you can find the updated configuration example for referer parser enrichment [here](https://docs.snowplowanalytics.com/docs/enriching-your-data/available-enrichments/referrer-parser-enrichment/).

PR #4283, opened 2020-02-10
Beam Enrich: extract into separate repo (closes #4282)

peel:
- [x] Needed for backporting framework version updates: https://github.com/snowplow/beam-enrich/issues/3
- [x] Description on gitflow branching: https://github.com/snowplow-incubator/engineering-resources/pull/15
---
Part of https://github.com/snowplow-incubator/engineering-resources/issues/8 effort.

PR #4312, opened 2020-04-06
Sqs buffer

lukeindykiewicz:
As this is not release PR I think it's enough to squash commits on merge to be aligned with 1:1:1 rule (is that correct?).

The changes to the collector that uses SQS as a buffer for load spikes. Is abandons retrying to Kinesis, but in stead of that it sends messages to sqs. The internal buffering is left as it was before. The mechanism tries to send last batch of messages to sqs as jvm shutdown hook (also left as it was before), so if this will not be successful there will be some messages lost.

SQS is a queue and it doesn't support key and value pairs (as kinesis does), so there is a hack to keep the keys. I wrote the comment is the code to explain it.


lukeindykiewicz:
Please assume it's green 🟢 , the only think that is failing is fixed in https://github.com/snowplow/snowplow/pull/4283/, so it should be fine :)

2-collectors/scala-stream-collector/kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -255,19 +287,73 @@ class KinesisSink private (
...
+          val b64Encoded = encode(msgWithKey)
+          new SendMessageBatchRequestEntry(UUID.randomUUID.toString, b64Encoded)
+      }
+      val MaxSqsBatchSize = 10
+      encoded.grouped(MaxSqsBatchSize).foreach { encodedGroup =>
+        val batchRequest =
+          new SendMessageBatchRequest()
+            .withQueueUrl(sqs.sqsBufferName)
+            .withEntries(encodedGroup.asJava)
+        Try {
```
peel, line 339: `Either.catchNonFatal`?

2-collectors/scala-stream-collector/kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -128,6 +142,29 @@ object KinesisSink {
...
+
+  def sqsBuffer(
+    sqsBufferName: Option[String],
+    provider: AWSCredentialsProvider,
+    region: String
+  ): Either[Throwable, Option[SqsClientAndName]] =
+    sqsBufferName match {
+      case Some(name) =>
+        for {
+          amazonSqs <- createSqsClient(provider, region)
```
peel, line 163: I'd use flatMap to make it more concise

2-collectors/scala-stream-collector/kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -255,19 +287,73 @@ class KinesisSink private (
...
+          ()
+        }.recover {
+          case e =>
+            log.error(s"Error sending to SQS queue(${sqs.sqsBufferName}): ${e.getMessage()}")
+        }
+      }
+    }
+
+  private def encode(bufMsg: ByteBuffer): String = {
+    val buffer = java.util.Base64.getEncoder.encode(bufMsg)
```
peel, line 353: This is a common practice to use underlying JVM locale for encoding and in the long run we probably should standardize upon a single charset. @oguzhanunlu has seen some misbehaviour caused by the kind of behavior.
Not expecting to do anything about it now, but to flag the issue for all to know.

2-collectors/scala-stream-collector/project/metals.sbt
```diff
@@ -0,0 +1,4 @@
+// DO NOT EDIT! This file is auto-generated.
+// This file enables sbt-bloop to create bloop config files.
+
+addSbtPlugin("ch.epfl.scala" % "sbt-bloop" % "1.4.0-RC1-105-118a551b")
```
peel, line 4: do we need that?

2-collectors/scala-stream-collector/kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
oguzhanunlu: Just to share my particular case, running old version of event recovery resulted in successfully recovered events where some of the events had fields containing `?` instead of expected letters, whereas other recovered events had expected letters. It was because our software relied on system default and we've run it across dataflow runners, where some runners couldn't encode letters like `é`.

I don't expect anything on this without team consensus on charset discussion but we should be aware of the possibility esp. when we have limited or no control over deployment platform.

2-collectors/scala-stream-collector/kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
lukeindykiewicz: Yeah, sounds good to leave fatal errors and recover only from non-fatal. Thanks!

2-collectors/scala-stream-collector/kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
lukeindykiewicz: Thanks, I had two lines in for comp in the beginning, so it's kind of a leftover. Now even `map` is enough inside! Pattern match is needed on the outside, as None should be successful.

2-collectors/scala-stream-collector/kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
lukeindykiewicz: Thanks, good point! What do you think about setting `utf-8` here and where decode happens in sqs2kinesis?

2-collectors/scala-stream-collector/project/metals.sbt
lukeindykiewicz: Sorry! Added to global git ignore, so it won't happen again!

2-collectors/scala-stream-collector/kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
oguzhanunlu: > about setting utf-8 here

I think we should stay as is w.r.t. charset and have the ability to configure charset if we don't have charset control over the deployment platform. (e.g. dataflow)

2-collectors/scala-stream-collector/kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -255,19 +286,76 @@ class KinesisSink private (
...
       }
     }
 
+  private def putToSqs(sqs: SqsClientAndName, batch: List[(ByteBuffer, String)]): Future[Unit] =
+    Future {
+      log.info(s"Writing ${batch.size} messages to SQS queue: ${sqs.sqsBufferName}")
+      val encoded = batch.map {
+        case (msg, key) =>
+          // The key is either IP or random UUID, so neither should contin `|`.
+          // Sqs doesn't support keys, so `|` is used as a separator. The same needs to be set up in
```
benjben, line 323: `|` is not hard coded and can be anything

lukeindykiewicz: It's a default separator for now, so unless someone overwrite it it defaults to `|`

2-collectors/scala-stream-collector/kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -148,11 +186,17 @@ class KinesisSink private (
...
-
   log.info("Creating thread pool of size " + kinesisConfig.threadPoolSize)
+  maybeSqs match {
+    case Some(sqs) =>
+      log.info(
+        s"SQS buffer for failed Kinesis stream '$streamName' is set up as: ${sqs.sqsBufferName}"
+      )
+    case None =>
+      log.warn(
+        s"No SQS buffer set up, all failed events from Kinesis stream '$streamName' will be dropped. (consider setting a SQS Buffer in config.hocon)"
```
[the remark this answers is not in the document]
lukeindykiewicz: For surge protections it's useless. For Kinesis outage it might help, but only for a short while, because for real traffic this will use all the collectors memory quickly (due to how retries are implemented at the moment at least). When collector dies all the events it had been buffering are lost (it tries to send it, but I assume the outage doesn't stop exactly at the moment when collector dies and invokes the shutdown hook). We had discussion on slack (mostly with Josh and João, but on open channels) to use other mechanisms, like saving to disk, but finally the approach with sqs was the winner.

2-collectors/scala-stream-collector/kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -20,53 +20,67 @@ import scala.collection.JavaConverters._
...
 import model._
+import scala.util.Try
+import java.util.UUID
+import com.snowplowanalytics.snowplow.collectors.scalastream.sinks.KinesisSink.SqsClientAndName
 
 /** KinesisSink companion object with factory method */
 object KinesisSink {
 
-  @volatile var shuttingDown = false
+  case class SqsClientAndName(sqsClient: AmazonSQS, sqsBufferName: String)
```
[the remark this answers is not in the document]
lukeindykiewicz: I'll check that

lukeindykiewicz:
> If Kinesis isn't available for a few minutes, and someone doesn't want to pay for SQS, the messages are directly lost, whereas today they are retried and can be inserted after Kinesis outage. So it seems to me that we should keep the exponential retry with a maximum delay and then insert to Kinesis in case of multiple failures, otherwise this seems like a regression compared with today's collector.

I don't think the `doesn't want to pay for SQS` is truly the argument here. Compare my comment https://github.com/snowplow/snowplow/pull/4312#discussion_r419428824 The cost of running the `surge protection` (the whole solution) should be close to zero when buffer is not needed. When events are actually going through sqs and sqs2kinesis it saves the events from being lost. Not losing client data is basically one of the core values for Snowplow (Yali's words paraphrased).

There is an ongoing discussion (with Josh and Yali) if the buffer should be hidden behind the feature flag, so it should be either the old code or the code from this PR. In case the feature flag is the way to go (it's most proabbable) I would close this PR and open new one.

In case we would go with feature flag it should allow clients to decide whether use it or not (pay for it or not).

2-collectors/scala-stream-collector/kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
lukeindykiewicz: I also have a better idea how to implement it, so I wold remove the whole separator away.

benjben: Great!

lukeindykiewicz:
Closing as newer version (containing also original code for retries) exists in https://github.com/snowplow/snowplow/pull/4346

PR #4327, opened 2020-04-21
Stream Enrich: add the possibility to override the DynamoDB endpoint

benjben:
This PR contains:
- possibility to override the DynamoDB endpoint for Kinesis checkpointing
- possibility to disable Cloudwatch, it seems that there are issues when using it with localstack

### DIFF
--- a/3-enrich/stream-enrich/examples/config.hocon.sample
+++ b/3-enrich/stream-enrich/examples/config.hocon.sample
@@ -69,6 +69,15 @@ enrich {
       # customEndpoint = {{kinesisEndpoint}}
       # customEndpoint = ${?ENRICH_STREAMS_SOURCE_SINK_CUSTOM_ENDPOINT}
 
+      ## Optional endpoint url configuration to override aws dyanomdb endpoints for Kinesis checkpoints lease table,
+      ## this can be used to specify local endpoints when using Localstack
+      # dynamodbCustomEndpoint = "http://localhost:4569"
+      # dynamodbCustomEndpoint = ${?ENRICH_DYNAMODB_CUSTOM_ENDPOINT}
+
+      # Optional override to disable cloudwatch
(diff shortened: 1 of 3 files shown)


3-enrich/stream-enrich/examples/config.hocon.sample
```diff
@@ -69,6 +69,13 @@ enrich {
       # customEndpoint = {{kinesisEndpoint}}
       # customEndpoint = ${?ENRICH_STREAMS_SOURCE_SINK_CUSTOM_ENDPOINT}
 
+      ## Optional endpoint url configuration to override aws dyanomdb endpoints for Kinesis checkpoints lease table,
+      ## this can be used to specify local endpoints when using Localstack
+      # dynamodbCustomEndpoint = "http://localhost:4569"
```
oguzhanunlu, line 74: Shouldn't we offer the ability to provide this field through an environment variable? Like any other field in this config template.

3-enrich/stream-enrich/examples/config.hocon.sample
```diff
@@ -69,6 +69,13 @@ enrich {
...
+
+      # Optional override to disable cloudwatch
+      # disableCloudWatch = true
```
oguzhanunlu, line 77: The question I asked for the other field applies here too

3-enrich/stream-enrich/examples/config.hocon.sample
benjben: It's already the case. Once the key `dynamodbCustomEndpoint` is read in the code it can be specified either with direct value e.g.
```
dynamodbCustomEndpoint = "http://localhost:4569"
```
or with an environment variable e.g.
```
dynamodbCustomEndpoint = ${?WHATEVER_ENV_VAR}
```

PR #4331, opened 2020-04-23
Fix NullPointerException IgluAdapter

benjben:


benjben:
@chuwy I added better unit testing. Given that the raw query string is parsed inside a `Loader` of enrich, I created a Thrift payload as the one created in SSC to use as input of the test. This logic opens the way to stronger unit testing in enrich.

chuwy:
Looks awesome!

PR #4371, opened 2020-08-05
Fix README dependency name

hrvolapeter:


benjben:
Thanks for the contribution @hrvolapeter !

Issue #4433, opened 2021-03-12
Restore CHANGELOG


alexanderdean:
In a recent tidy-up of this repo we lost the `CHANGELOG` in the root.

I think this is a shame, for two reasons:

1. We lost a visible historical record ("it's in the git history" doesn't count as visible) of the evolution of our core IP over the past 8 years
2. We have lost a prominent place to log the ongoing changes we are making to this repo (for example @adatzer's great work on discoverability via git submodule updates)

I propose:

* Resurrecting the historic CHANGELOG
* Backfilling entries since its deletion
* Adding new entries as we make updates to this repo

For new entries in the CHANGELOG, I would propose simply datestamping them given that they don't fall under specific release numbers any more (there is no global Snowplow release version at the level of this repo currently - we don't offer LTS / blessed bundles to OSS yet).

paulboocock:
We should do this (and automate it) when we automate the updating of the submodules, which we've just started working on (cc @adatzer). 

I'd suggest introducing a CHANGELOG.json too as this will make autogenerating the CHANGELOG.md far easier and more reliable.

benjben:
In the tidy-up of this repo I had deleted the `CHANGELOG` on purpose because I failed to see the need for it in a post world umbrella releases world, but I can now see the value.

I think that it would be great to automatically update it as such : unless I'm missing something, a submodule will get updated only when new commits are available on the `master` branch of the child repo. This happens only when there has been a new release of a component. In this case the latest commit should always be named `Prepare for release x.y.x` according to our workflow. We could then take this version `x.y.z` and add one line to the `CHANGELOG` when this happens, like below :

```
2021-03-08 - snowplow/stream-collector - Release 2.2.0
2020-11-20 - snowplow/enrich - Release 1.4.2
2020-08-27 - snowplow/iglu-scala-client - Release 1.0.2
2020-07-22 - snowplow-incubator/snowplow-badrows - Release 2.1.0
```

I think that `snowplow/snowplow` should still get all the commits all the child repos and not only the `Prepare for release` commits.

How does that sound?

adatzer:
Indeed @benjben that would be definitely possible. 
There is the concern of it just repeating the commits history without a wrapping story, so maybe we could just restore it for now and reconsider its automation in relation to a possible "release" style? cc @paulboocock

benjben:
[...]

PR #4436, opened 2021-03-19
Issue/update submodules automate

adatzer:
- As a test, the `Submodules_update` workflow can be seen in this temporary fork [here](https://github.com/adatzer/snowplow/actions), where the github-actions bot made the same bump [commits](https://github.com/adatzer/snowplow/commits/master) we made manually here, and also the expected ones after adding the `data-models` submodule there too. 
- The workflow can also be run manually, for the case we want an update before the next scheduled run.


.scripts/lib.sh
```diff
@@ -0,0 +1,92 @@
...
+    git merge --ff "${__bump_commit}"
+
+    # 2. go toplevel
+    cd "${__top_level}"
+
+    # 3. get repo name for commit message
+    repo_url=`git submodule--helper config submodule."${__mod_name}".url`
+    [ -z "${repo_url}" ] && die "Submodule url not found for ${__mod_name}"
+    repo_name=`basename "${repo_url}" .git`
+    commit_msg="${repo_name}: Bump to ${__bump_tag}"
```
benjben, line 83: I would replace `Bump to` with `Release`

.scripts/lib.sh
```diff
@@ -0,0 +1,92 @@
...
+            [[ "${__tag}" = $tag_pattern ]] && echo "${__tag#v}" || return 1
+            ;;
+        4-storage/snowplow-rdb-loader)
+            tag_pattern='r+([0-9])'
+            [[ "${__tag}" = $tag_pattern ]] && echo "${__tag}" || return 1
+            ;;
+        5-data-modeling/data-models)
+            tag_pattern='**/+([0-9]).+([0-9]).+([0-9])'
+            [[ "${__tag}" = $tag_pattern ]] && echo "${__tag}" || return 1
+            ;;
```
benjben, line 56: Great that you checked each individual repo. We need to unify our tagging style across all the repos.

benjben (commented):
This is really great @adatzer !

I think that we are just missing the automatic update of `CHANGELOG`. You can see a proposal about what to put inside in [this comment](https://github.com/snowplow/snowplow/issues/4433#issuecomment-802965797).

.scripts/lib.sh
adatzer: Thank you for reviewing @benjben !
The reason i did it this way was mainly to stay consinstent with the commits' messages of the manual updates. I can certainly change it to `Release`. Do you think it would be preferred from the point of view of the `snowplow/snowplow` repository?

benjben: I think that `Release` sends a stronger message than `Bump`, but that might be just a matter of taste.

.scripts/lib.sh
```diff
@@ -0,0 +1,92 @@
...
+function is_release_tag() (
+    [ "$#" -eq 2 ] || die "2 arguments required, $# provided to 'func:is_release_tag'"
+    local __tag="$1"
+    local __module="$2"
+
+    # For extended glob patterns and ensure [0-9] range
+    shopt -s extglob
+    LC_ALL=C
+
+    case "${__module}" in
```
paulboocock, line 36: I think we will need to update this to add a JavaScript tracker specific rule to it. 

With v3, we will end up with a `3.0.0` tag but also a tag for each package (that will happen automatically by our publishing tool) like `@snowplow/browser-tracker_v3.0.0`. We only need the 3.0.0 style tags here, not the NPM package tags.
If you'd like an example, it will end up something like this repo: https://github.com/paulboocock/rush-playground

paulboocock (approved):

paulboocock (approved):

PR #4554, opened 2022-09-08
Fix docs links for enrich module

colmsnowplow:
PR to address broken links flagged here: https://discourse.snowplow.io/t/404-setup-validation-and-enrich/7407/6

### DIFF
--- a/3-enrich/README.md
+++ b/3-enrich/README.md
@@ -15,6 +15,6 @@ They are both part on [enrich][enrich] project.
 
 [storage]: https://github.com/snowplow/snowplow/tree/master/4-storage
 
-[docs-beam-enrich]: https://docs.snowplowanalytics.com/docs/setup-snowplow-on-gcp/setup-validation-and-enrich-beam-enrich/
-[docs-stream-enrich]: https://docs.snowplowanalytics.com/docs/setup-snowplow-on-aws/setup-validation-and-enrich/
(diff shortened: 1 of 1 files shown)


3-enrich/README.md
```diff
@@ -15,6 +15,6 @@ They are both part on [enrich][enrich] project.
 
 [storage]: https://github.com/snowplow/snowplow/tree/master/4-storage
 
-[docs-beam-enrich]: https://docs.snowplowanalytics.com/docs/setup-snowplow-on-gcp/setup-validation-and-enrich-beam-enrich/
-[docs-stream-enrich]: https://docs.snowplowanalytics.com/docs/setup-snowplow-on-aws/setup-validation-and-enrich/
+[docs-beam-enrich]: https://docs.snowplow.io/docs/getting-started-on-snowplow-open-source/setup-snowplow-on-gcp/setup-validation-and-enrich/
```
stanch, line 18: I would also fix the link names, not that it changes the resulting text, but could be confusing when we edit it.

stanch: (It should not say `beam-enrich`)

stanch: Actually hold on, we need to fix the text as well! It should be enrich-kinesis and enrich-pubsub, not stream-enrich and beam-enrich.

stanch (changes requested):
Mention the correct modules in the text

3-enrich/README.md
colmsnowplow: Oh true, good catch. One sec

stanch (approved):

